PR #1, opened 2016-09-02
Use rails engine implementation of models.

eugenk:
This uses the models defined in the rails engine at https://github.com/ontohub/ontohub-models/pull/1.

Right before merging, I need to merge the above pull request and then update the Gemfile to use the `master` branch of https://github.com/ontohub/ontohub-models.

It moves the monkey-patch of `sequel-rails` from this repository to https://github.com/ontohub/ontohub-models.


Gemfile
```diff
@@ -11,7 +11,7 @@ gem 'puma', '~> 3.0'
 # Use Rack CORS for handling Cross-Origin Resource Sharing (CORS), making cross-origin AJAX possible
 gem 'rack-cors'
 
-gem 'ontohub-models', github: 'ontohub/ontohub-models', branch: 'master', require: false
+gem 'ontohub-models', github: 'ontohub/ontohub-models', branch: 'create_models_as_rails_engine'
```
phyrog: Shouldn't it be the `master` branch?

eugenk: It should, but only after the master branch is up to date. I need to change this line after merging ontohub-models/pulls/1 and before merging this one. Or maybe it's better to specify a commit?

phyrog: The exact commit is pinned in the `Gemfile.lock`, so `master` should be OK.

phyrog:
Looks good. :+1:

PR #3, opened 2016-09-08
Remove direct dependency: sequel-rails.

eugenk:
This is not needed because it is already an indirect dependency (though ontohub-models). It is a leftover of the non-rails-engine structure of ontohub-models.


phyrog:
:+1:

PR #4, opened 2016-09-12
Fix rubocop config

eugenk:
Some cops have been deprecated since our last change. This commit uses
the current default configuration and changes it according to our style
guide.


ebolloff:
:+1:

PR #5, opened 2016-09-12
Create first controller

eugenk:
This creates the first [JSON API 1.0](http://jsonapi.org/format/1.0/) compatible controller with its specs.


app/controllers/v2/repositories_controller.rb
```diff
@@ -0,0 +1,69 @@
...
+      @collection = Repository.all
+      render json: @collection, each_serializer: V2::RepositorySerializer
+    end
+
+    def show
+      retrieve_resource
+      render_resource
+    end
+
+    def create
```
phyrog: We should probably find a way to easily adhere to the return codes specified in the jsonapi.org spec (here: http://jsonapi.org/format/#crud-creating-responses). Do you know of an good way to share this status logic between all the controllers we will write?

This of course applies to all other actions as well.

app/controllers/v2/resources_controller/resources_helpers.rb
```diff
@@ -0,0 +1,96 @@
...
+        end
+
+        # This will try to infer the model class. Do not call it directly from
+        # the controller. It is supposed to be called during loading of the
+        # controller only.
+        def infer_resource_class
+          klass = to_s.split(':').last.sub(/Controller\z/, '').singularize.
+            constantize
+          resource_class(klass)
+        rescue NameError
```
eugenk: At this point, we have to suppress it. See the code comment in the next line.

app/controllers/v2/repositories_controller.rb
eugenk: The only way I can think of is to use a handy superclass that defines all this for the typical case. I created a `ResourceController` similar to the basic idea of [inherited resources](https://github.com/josevalim/inherited_resources) but with much less functionality/magic. It gives us a simple way to stay DRY and a place where we can lookup defaults if we need to.

Gemfile
```diff
@@ -13,7 +13,10 @@ gem 'puma', '~> 3.0'
 # cross-origin AJAX possible
 gem 'rack-cors'
 
-gem 'ontohub-models', github: 'ontohub/ontohub-models', branch: 'master'
+gem 'ontohub-models', github: 'ontohub/ontohub-models',
+                      branch: 'allow_to_use_active_model_serializers'
```
eugenk: Done by commit amend.

eugenk:
> I think we should wait with merging this until ontohub/ontohub-models#6 is finished, so we don't need to extend this later.

No, we should not. https://github.com/ontohub/ontohub-models/issues/6 adds more requirements to the repository (e.g. namespace required) which are not part of this pull request. This should be done in #6.

This is also the reason why I pinned the previous commit of ontohub-models in the Gemfile.lock.

app/controllers/v2/repositories_controller.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+module V2
+  # Handles all requests for repository CRUD operations
+  class RepositoriesController < ResourcesController
+    find_param :slug
+    permitted_params :name, :description
```
phyrog: I believe this also allows the name to be edited. Is that wanted? If not, we need to forbid that in the `update` action.

PR #13, opened 2016-11-04
Remove rspec task from rakefile

### LINKED ISSUE #12: RSpec is run twice when using rake
```
RSpec::Core::RakeTask.new(:spec)
```

should be removed from the Rakefile.

phyrog:
Should fix #12. Rspec-rails already defines a task `spec` and sets it as default, so when we define one as well, both are run. Unfortunately this already defined task can not be run by using `rake spec`, so the only way to run the specs is by running `rake`.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -8,10 +8,4 @@ require_relative 'config/application'
 
 Rails.application.load_tasks
 
-begin
-  require 'rspec/core/rake_task'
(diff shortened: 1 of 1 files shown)


phyrog:
Failing builds are caused by #14

ebolloff:
:cat2:

PR #22, opened 2016-11-07
21 repository controller returns repositories of all namespaces

### LINKED ISSUE #21: repository controller returns repositories of all namespaces
Given there are two namespaces with repositories
When I call the API to get the repositories of one of the namespaces (e.g. `/namespaces/xyz/repositories`)
Then all repositories of both namespaces will be included

eugenk:
This shall fix #21.
It also changes the routes to avoid doubled namespace slug in the repository URL.
Furthermore, it adds a test for a failing update due to failing validations and it adds seeds.

### DIFF
--- a/app/serializers/v2/repository_serializer.rb
+++ b/app/serializers/v2/repository_serializer.rb
@@ -17,12 +17,11 @@ module V2
 
     link :self do
       url_for(controller: 'v2/repositories', action: 'show',
-              namespace_slug: object.namespace.to_param,
-              slug: object.slug)
+              slug: object.to_param)
     end
 
     def id
-      object.slug
+      object.to_param
     end
   end
 end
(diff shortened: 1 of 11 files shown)


eugenk:
This should be fixed with 043a223 and 662532c.

app/serializers/v2/repository_serializer.rb
```diff
@@ -16,8 +16,10 @@ class RepositorySerializer < ApplicationSerializer
     end
 
     link :self do
```
0robustus1: Just one suggestion: Consider making the urls part of the model (generally). That way a model 'owns' and 'knows' it's location on the web, which is one of the key-informations. It also removes the need to potentially build urls in multiple places.

eugenk: Thanks! We plan to do this in a later step, when we introduce Loc/Ids again.

PR #23, opened 2016-11-10
Update gems

eugenk:
This keeps the gems up to date.

* [x] This needs to be rebased on #22 in order to be green.


Gemfile.lock
```diff
@@ -94,12 +95,12 @@ GEM
       warden (~> 1.2.3)
     diff-lcs (1.2.5)
     docile (1.1.5)
-    elasticsearch (2.0.0)
-      elasticsearch-api (= 2.0.0)
-      elasticsearch-transport (= 2.0.0)
-    elasticsearch-api (2.0.0)
+    elasticsearch (5.0.0)
```
eugenk: I think the gem version is now aligned with the application version of Elasticsearch.

PR #29, opened 2016-11-18
28 add settings for host

### LINKED ISSUE #28: Add Settings for Host
Currently, `url_for` always returns `http://localhost:3000/...`. This must be configured to use a value that is configurable.

eugenk:
This shall fix #28.


app/serializers/v2/namespace_serializer.rb
```diff
@@ -6,15 +6,17 @@ class NamespaceSerializer < ApplicationSerializer
     attribute :name
 
     link :self do
-      url_for(controller: 'v2/namespaces', action: 'show',
-              slug: object.to_param)
+      path = url_for(controller: 'v2/namespaces', action: 'show',
+                     slug: object.to_param, only_path: true)
+      [Settings.server_url, path].join('/')
```
phyrog: These will be simplified later by models knowing their own path; For the meantime, this seems to be the easiest solution, even if we're duplicating some code.

PR #27, opened 2016-11-18
20 create dummy search controller

eugenk:
This shall fix #20. The output of `http -j :3000/search` looks like this:

```json
{
    "data": {
        "attributes": {
            "repositories_count": 4,
            "results_count": 6,
            "users_count": 2
        },
        "id": "searchresult",
        "relationships": {
            "repositories": {
                "data": [
                    {
                        "id": "ada/repo1",
                        "type": "repositories"
                    },
                    {
                        "id": "bob/repo2",
                        "type": "repositories"
                    },
                    {
                        "id": "ada/repo3",
                        "type": "repositories"
                    },
                    {
                        "id": "bob/repo4",
                        "type": "repositories"
                    }
                ]
            },
            "users": {
                "data": [
                    {
                        "id": "ada",
                        "type": "users"
                    },
                    {
                        "id": "bob",
                        "type": "users"
                    }
                ]
            }
        },
        "type": "search_results"
    },
    "jsonapi": {
        "version": "1.0"
    }
}
```

And the output of `http -j :3000/users/ada` looks like this
```json
{
    "data": {
        "attributes": {
            "display_name": null,
            "email": null,
            "name": "ada"
        },
        "id": "ada",
        "links": {
            "self": "http://localhost:3000/users/ada"
        },
        "relationships": {
            "teams": {
                "data": []
            }
        },
        "type": "users"
    },
    "jsonapi": {
        "version": "1.0"
    }
}
```

PR #32, opened 2016-11-23
31 add api documentation

### LINKED ISSUE #31: Add API documentation
Although there are tests and JSON schema definitions, we could use a human-readable API documentation. I suggest to create a doc/api directory and put API documentation there.

eugenk:
This shall fix #31.

It also contains the current API additions from #27. If the API needs to be changed in #27 and this gets merged before #27, I will change the corresponding documentation in #27.


eugenk:
I rebased on master and I fixed the repository/namespace id as well as the create command in the last commit. If this pull requests becomes approved, I will squash the last two commits before merging.

PR #43, opened 2016-12-06
42 add organizationscontroller

eugenk:
This shall fix #42.

Todo:
* [x] Update Gemfile/models properly after https://github.com/ontohub/ontohub-models/pull/40 is merged.

PR #45, opened 2016-12-07
37 add short routes for namespace and repository

### LINKED ISSUE #37: Add Short Routes for Namespace and Repository
Add routes for `/:namespace_slug` and `/:namespace_slug/:repository_slug`.

Put the old namespaced routes in a scope `endpoints`.

eugenk:
This shall fix #37. This is based on #43. The first new commit is "Rename module to DSL. ea0f19c"

It first checks of which class the requested object is and then instantiates the corresponding controller to call the corresponding actions. The params must be changed to allow the target controller to work properly.

Todo:
* [ ] Rebase on master as soon as #43 is merged.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -1,4 +1,5 @@
 # frozen_string_literal: true
+# rubocop:disable Metrics/BlockLength
 
 Rails.application.routes.draw do
   scope format: false do
@@ -30,5 +31,10 @@ Rails.application.routes.draw do
 
     get 'search', controller: 'v2/search', action: 'search'
     get 'version', controller: 'v2/version', action: 'show'
+
+    get '/:slug',
+      controller: 'v2/organizational_units',
+      action: :show,
+      as: :organizational_unit
   end
 end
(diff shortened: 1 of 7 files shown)


config/routes.rb
```diff
@@ -1,29 +1,42 @@
 # frozen_string_literal: true
+# rubocop:disable Metrics/BlockLength
```
eugenk: Locally, this is needed. Does houndci use a different configuration?

config/routes.rb
```diff
@@ -28,6 +29,14 @@
       param: :slug,
       constraints: {slug: %r{[^/]+/[^/]+}}
 
-    get 'search', controller: 'v2/search', action: 'search'
+    get '/:slug',
+      controller: 'v2/organizational_units',
```
ebolloff: See comment from @phyrog in #43.

eugenk: In what way is his comment associated with this line?

phyrog: It's related to this line, since this is the route for `organizational_units`. To integrate the frontent better, another route `/organizational_units/:slug` would be nice. It should do the same thing as the route above.

PR #51, opened 2017-01-06
Update to Ruby 2.4.0

ebolloff:
It'll fail until the `json` gem is compatible with ruby 2.4.0


ebolloff:
We'll wait for Rails, because we get a deprecation warning locally.

ebolloff:
Update: We're still waiting.

PR #59, opened 2017-01-24
Return all organizational_units in search API

### LINKED ISSUE #57: Search serializer should return merged users and organizations
They both share a common namespace anyway. Searching for either one alone can be done via a client side filter or with a filtering query parameter.

phyrog:
This fixes #57.


phyrog:
Coverage decreased because we have fewer overall lines of code.

PR #60, opened 2017-01-24
Don't run version check in test env

### LINKED ISSUE #55: Version check fails on Travis
The current version check introduced in #47 uses `git describe --long --tags` to get the latest git tag. Since Travis only clones the last 50 commits and we now have more than 50 commits on some new branches, the tag is not cloned anymore and thus the version check fails.

The main problem is, that the command is run at startup time and not during the tests (during which the code is stubbed). So basically, if we're in test mode, we don't want to run the command at all during startup.

phyrog:
This fixes #55.

### DIFF
--- a/app/models/version.rb
+++ b/app/models/version.rb
@@ -35,23 +35,29 @@ class Version < ActiveModelSerializers::Model
 
     private
 
+    # :nocov:
     def production?
       Rails.env.production?
     end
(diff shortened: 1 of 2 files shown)


phyrog:
The now untested LOC were only accidentally implicitly tested before because they were run when the application started. They are actually stubbed in the tests.

PR #69, opened 2017-02-01
Don't add devise routes if running certain rake tasks

### LINKED ISSUE #67: Solve devise/migrate issue
```
rails db:drop
rails db:create
rails db:migrate
```
Fails because of the `devise_for` in the routes. Fix it.

phyrog:
This fixes #67. Initially I wanted to change sequel-rails to not use the environment in the `db:` tasks at all, but of course the database config is needed. So we would have to somehow load the environment without the routes.

The current solution is easier and robust: Checking if we're running `rake` by checking if `Rake` is defined and if the current top level tasks include `db:create`, `db:migrate` or `db:drop`. In these cases the devise routes are not added. In all other cases the routes are still there, so we can run `rails routes` and see the devise routes as well.

If we encounter other rake tasks that fail because of devise being loaded in the routes, we can simply add them to the blacklist.

How to test:

```
rails db:drop (if the database already exists, else skip)
rails db:create
rails db:migrate
rails db:routes (see that the devise routes are there)
```

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -1,10 +1,24 @@
 # frozen_string_literal: true
 # rubocop:disable Metrics/BlockLength
 
+# :nocov:
+def rake_task?(tasks = [])
+  defined?(Rake) && tasks.any? do |task|
+    Rake.application.top_level_tasks.include?(task)
+  end
+end
+# :nocov:
+
 Rails.application.routes.draw do
   scope format: false, defaults: {format: :json} do
     root to: 'v2/search#search'
-    devise_for :users, controllers: {sessions: 'v2/users/sessions'}
+
+    # We exclude the devise routes from these rake tasks because they load the
+    # models. When loading the models, the database needs to exist, or else it
+    # throws an error.
+    unless rake_task?(%w(db:create db:migrate db:drop))
+      devise_for :users, controllers: {sessions: 'v2/users/sessions'}
+    end
     resources :organizations,
       controller: 'v2/organizations',
       only: :show,


config/routes.rb
```diff
@@ -1,10 +1,18 @@
...
+  (defined? Rake) && tasks.any? do |task|
+    Rake.application.top_level_tasks.include?(task)
+  end
+end
+
 Rails.application.routes.draw do
   scope format: false, defaults: {format: :json} do
     root to: 'v2/search#search'
-    devise_for :users, controllers: {sessions: 'v2/users/sessions'}
+    unless rake_task?(%w(db:create db:migrate db:drop))
```
eugenk: Please add an inline comment describing the reason for this exception.

config/routes.rb
```diff
@@ -1,10 +1,18 @@
 # frozen_string_literal: true
 # rubocop:disable Metrics/BlockLength
 
+def rake_task?(tasks = [])
+  (defined? Rake) && tasks.any? do |task|
```
eugenk: Code style: Please use parentheses like in `defined?(Rake)`.

phyrog: Done

config/routes.rb
phyrog: Done

config/routes.rb
```diff
@@ -1,10 +1,22 @@
 # frozen_string_literal: true
 # rubocop:disable Metrics/BlockLength
 
+def rake_task?(tasks = [])
+  defined?(Rake) && tasks.any? do |task|
+    Rake.application.top_level_tasks.include?(task)
```
eugenk: Oh and please wrap this line in a `# :nocov:` environment since it's impossible to (reasonably) execute it in a test.

phyrog: Just the line or the whole method?

eugenk: The whole method is okay as well.

phyrog: Ok, done

Issue #61, opened 2017-01-25
Auto generate API documentation from JSON Schema


phyrog:
I just discovered [doca](https://github.com/cloudflare/doca), which is a tool developed by CloudFlare, that generates HTML from JSON Schema files. Here is an introductory blog post, that shows how this works: https://blog.cloudflare.com/cloudflares-json-powered-documentation-generator/

Also: [CloudFlares own API documentation](https://api.cloudflare.com/) generated with doca

Might be worth looking into, given that we already have JSON Schema files for the API tests.

phyrog:
For this to work we need to annotate our json schemas with more data (descriptions, example values, constraints).

PR #72, opened 2017-02-02
71 structure json schema files

### LINKED ISSUE #71: Structure json schema files
The JSON schema files that document the API need to be structured strictly for https://github.com/ontohub/ontohub-frontend/issues/41 to work.

I suggest the following file names and directory tree:
* `spec/support/api/schemas/`: The JSON Schema directory
  * `controllers/<controller_name>/`: All the actions of the controllers are defined in this place. There is one file per action, for example, `v2/repositories/get_show.json`. These may contain `$ref`s to files of the following bullet.
  * `models/`: Definitions of the models with their attributes, links and relationships. For instance, a `repository_model.json` would be here. These will be needed in the frontend to generate models there. The relationships themselves are just `$ref`s to files of the following bullet.
  * `relationships/`: Definitions of the structure that the models have inside a `relationships` object. For example, the definition of `properties.relationships.owner` of a `repository` would be placed in the `organizational_unit_relationship.json` in this directory.

Every file should have the following header:
```json
{
  "$schema": "http://json-schema.org/draft-04/schema#",
  "title": "<title for the documentation>",
  "description": "<description fot the documentation>",

  // the actual content of this schema goes here...
}
```
Note that there is no `id` of the schema itself any more.
Also, schemas that are referenced via `$ref` are defined in the root object and not nested in a `definitions` object.

eugenk:
This shall fix #71.

It makes the structure also compatible with doca (see #61).


spec/support/json_schema_matcher.rb
```diff
@@ -1,14 +1,56 @@
 # frozen_string_literal: true
 
-RSpec::Matchers.
-  define :match_response_schema do |api_version, schema|
+# Finds and ransforms constant V2::UsersController to string "v2/users".
```
ebolloff: Do you mean `transforms`?

eugenk: Fixed by commit amend.

PR #62, opened 2017-01-27
53 login with jwt and devise

ebolloff:
This should fix #53

PR #74, opened 2017-02-03
Add more tests for JWT

phyrog:
Increases the test coverage of `config/initializers/core_extensions/devise/strategies/json_web_token.rb` and `app/helpers/jwt_wrapper.rb` to 100%.

### DIFF
--- a/config/initializers/core_extensions/devise/strategies/json_web_token.rb
+++ b/config/initializers/core_extensions/devise/strategies/json_web_token.rb
@@ -5,7 +5,7 @@ module Devise
     # strategy for login with jwt
     class JsonWebToken < Base
       def valid?
-        return false unless request.headers['Authorization'].present?
+        return false unless request.headers['HTTP_AUTHORIZATION'].present?
 
         strategy, _token = strategy_and_token
         (strategy || '').casecmp 'bearer'
@@ -20,7 +20,7 @@ module Devise
       protected
 
       def strategy_and_token
-        @strategy_and_token ||= request.headers['Authorization'].split(' ')
+        @strategy_and_token ||= request.headers['HTTP_AUTHORIZATION'].split(' ')
       end
 
       def claims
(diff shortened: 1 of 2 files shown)


config/initializers/core_extensions/devise/strategies/json_web_token.rb
```diff
@@ -5,7 +5,7 @@ module Strategies
     # strategy for login with jwt
     class JsonWebToken < Base
       def valid?
-        return false unless request.headers['Authorization'].present?
+        return false unless request.headers['HTTP_AUTHORIZATION'].present?
```
phyrog: Not sure if this change is necessary (probably not) or what the rails convention is for using `Authorization` vs `HTTP_AUTHORIZATION`.

spec/controllers/v2/users/sessions_controller_spec.rb
```diff
@@ -33,5 +33,31 @@
         it { expect(response_hash['error']).not_to be_empty }
       end
     end
+
```
ebolloff: I wanted to do that in #62, but we were not sure, if this is the right place for it. @eugenk what would you suggest?

config/initializers/core_extensions/devise/strategies/json_web_token.rb
phyrog: `HTTP_AUTHORIZATION` is case-insensitive against the client header, so I would prefer to use it instead of `Authorization`.

ebolloff: I would prefer `HTTP_AUTHORIZATION`, too. 👍

phyrog:
I think coverall says the coverage decreased because I removed a line in an amend, so the coverage decreased in comparison to the old commit because there is now one line fewer. :roll_eyes:

spec/controllers/v2/users/sessions_controller_spec.rb
phyrog: Well, since we can sign in with a token as well (i.e. generate a new token), we should test that code path as well IMHO.

ebolloff: @eugenk says, it's okay.

PR #114, opened 2017-05-12
#112 Login broken

### LINKED ISSUE #112: Login broken
Start the server and try
```
TOKEN=`http POST :3000/users/sign_in user:='{"name": "ada", "password": "changeme"}' | ruby -e 'puts $stdin.read.match(/"token":\s*"([^"]+)"/)[1]'`
```
and
```
http -j :3000/ada "Authorization: Bearer ${TOKEN}"
```
This should respond with `200`, which it does.
```
http -j :3000/ada "Authorization: Bearer ${TOKEN}asdf"
```
This should respond with a login error, but it responds with `200`.
Also, when you try to login with a non-existant user, the user gets a token which isn't allowed:
```
http POST :3000/users/sign_in user:='{"name": "NotAUser", "password": "changeme"}'
```

phyrog:
Fixes #112. This PR removes the devise guest strategy (which was just a hack to begin with). The method `current_user` does the same thing as `authenticate_user!`, except it sets the `current_user` to `nil` instead of failing completely. The guest strategy also caused one of the problems described in #112.


phyrog:
Coverage decreased because of less overall code.

PR #116, opened 2017-05-15
Update rails to 5.1.1

ebolloff:


ebolloff:
Travis fails because https://github.com/ontohub/ontohub-models/pull/70 fails.

PR #120, opened 2017-05-29
104 outsource the git layer to gitlab git

### LINKED ISSUE #104: Outsource the git layer to gitlab_git
The git layer may be interesting for other people as well. Since gitlab does not maintain gitlab_git any more, we should move the library functionality of our git layer to our fork of the gitlab_git gem.

eugenk:
This shall fix #104 and depends on https://github.com/ontohub/gitlab_git/pull/2.

It also adds a `TreesController#multiaction` to allow multiple changes in a single commit.


eugenk:
Changes since last review:
* rebased on master
  * resolved the conflict on app/models/[mutli_]blob.rb: use `display_name` instead of `real_name`
* add commit to use gitlab_git's master

eugenk:
I don't know why hound is complaining. My tests pass, so there can't be a syntax error.

PR #121, opened 2017-05-30
Replace coveralls with codecov

phyrog:


phyrog:
Not sure why this is breaking on Travis while it's working for me locally.

eugenk:
I rebased this on master and added a commit that adds `require: false` to codecov in the Gemfile.

PR #136, opened 2017-06-18
Create graphql query for organizational units

### LINKED ISSUE #133: Add GraphQL query for OrganizationalUnit
## Types:

```graphql
input UserUpdate {
  password: String
  displayName: String
  email: String
}

input NewOrganization {
  description: String
  displayName: String
}

input OrganizationUpdate {
  description: String
  displayName: String
}

type Organization implements OrganizationalUnit {
  description: String
  displayName: String
  id: ID!
  members(
    limit: Int = 20
    skip: Int = 0
  ): [User]!
}

interface OrganizationalUnit {
  displayName: String
  id: ID!
}

type User implements OrganizationalUnit {
  displayName: String
  email: String
  emailHash: String!
  id: ID!
  organizations(
    limit: Int = 20
    skip: Int = 0
  ): [Organization]!
}
```

## Queries

```graphql
organizationalUnit(id: ID!): OrganizationalUnit
```

## Mutations

```graphql
createOrganization(data: NewOrganization): Organization
saveOrganization(id: ID!, data: OrganizationUpdate): Organization
deleteOrganization(id: ID!): Boolean!

saveAccount(data: UserUpdate!, password: String!): User
deleteAccount(password: String!): Boolean!
```

phyrog:
Fixes #133.

After changing the graphql schema (e.g. by adding fields, changing field types, removing fields) `rake graphql:write` must be run to regenerate the file `/spec/support/schema.graphql`. Otherwise the corresponding test will fail. This test is there to spot API changes during code reviews more easily.

**Types**
- [x] UserChangeset
- [x] NewOrganization
- [x] OrganizationChangeset
- [x] Organization
- [x] OrganizationalUnit
- [x] User

**Queries**
- [x] organizationalUnit

**Mutations**
- [x] createOrganization
- [x] saveOrganization
- [x] deleteOrganization
- [x] saveAccount
- [x] deleteAccount


phyrog:
I'm very open to comments on the tests/test structure (what to test in general, where to test what, etc), since I haven't found any guidelines on that for GraphQL.

phyrog:
Test currently failing because of ontohub/sequel-devise#1

app/graphql/instrumenters/resource_instrumenter.rb
```diff
@@ -0,0 +1,30 @@
...
+
+GraphQL::Field.accepts_definitions resource: GraphQL::Define.assign_metadata_key(:resource)
+
+module Instrumenters
+  # Allows specifying a `resource` proc on fields, whose result will be passed
+  # to the resolve function of the field as the first argument.
+  # Authorization should be performed after this step, so the object can be
+  # properly authorized
+  class ResourceInstrumenter
+    def instrument(_type, field)
```
phyrog: Fixed in 1fefe9e7d65cb65f92fc03f51befa59e7ec55a67

spec/graphql/mutations/delete_account_mutation_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+    let(:variables) { {'password' => 'changemeow'} }
+    subject { result }
+
+    it 'does not delete the account' do
+      result
+      expect(User.find(id: user.id)).to_not be(nil)
+    end
+  end
+
+  context 'User does not exist' do
```
eugenk: How about `'Current user does not exist'`?

spec/graphql/mutations/create_organization_mutation_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+    before do
+      create :organization, name: organization_data['name']
+    end
+
+    subject { result }
+
+    it 'returns an error' do
+      expect(subject['data']['createOrganization']).to be(nil)
+      expect(subject['errors']).to include(
+        include('message' => 'name is already taken')
```
eugenk: Why the double `include`?

The same question applies to the other specs.

spec/graphql/mutations/create_organization_mutation_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+        members {
+          id
+        }
+      }
+    }
+    QUERY
+  end
+
+  context 'Successful creation' do
+    subject { result }
```
eugenk: Why don't you put the `subject` outside the `context`?

The same question applies to the other specs.

spec/graphql/mutations/delete_account_mutation_spec.rb
```diff
@@ -0,0 +1,57 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+RSpec.describe 'deleteAccount mutation' do
+  let!(:user) { create :user, password: 'changemenow' }
```
eugenk: Please use
```ruby
let!(:user) { create :user }
let(:password) { user.password }
```
and use the password variable everywhere.

The same applies to the other specs.

spec/graphql/mutations/save_account_mutation_spec.rb
```diff
@@ -0,0 +1,76 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+RSpec.describe 'saveAccount mutation' do
+  let!(:user) { create :user, password: 'changemenow' }
```
eugenk: Please use
```ruby
let!(:user) { create :user }
let(:password) { user.password }
let(:new_password) { "changed-#{password}" }
```
and use the password variable everywhere.

spec/graphql/mutations/save_account_mutation_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+        'id' => user.slug,
+        'displayName' => user_data['displayName'],
+        'email' => user.email,
+        'unconfirmedEmail' => user_data['email']
+      )
+    end
+  end
+
+  context 'Incorrect password' do
+    let(:variables) { {'data' => user_data, 'password' => 'changemeow'} }
```
eugenk: `'password' => "bad-#{password}"`

spec/graphql/mutations/save_account_mutation_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+    subject { result }
+
+    it 'returns the new account fields' do
+      expect(subject['data']['saveAccount']).to include(
+        'id' => user.slug,
+        'displayName' => user_data['displayName'],
+        'email' => user.email,
+        'unconfirmedEmail' => user_data['email']
+      )
+    end
```
eugenk: Please add a spec that tests the new password.

spec/support/schema.graphql
```diff
@@ -0,0 +1,95 @@
+type Mutation {
+  createOrganization(data: NewOrganization!): Organization
+  deleteAccount(password: String!): Boolean
+  deleteOrganization(id: ID!): Boolean
+  saveAccount(data: UserChangeset!, password: String!): User
+  saveOrganization(id: ID!, data: OrganizationChangeset!): Organization
+}
```
eugenk: I don't have the picture of the whiteboard with our notes of last week. Are these the types we agreed upon? They look reasonable.

app/controllers/v2/application_controller.rb
```diff
@@ -5,6 +5,5 @@ module V2
   class ApplicationController < ActionController::API
     # Accept and generate JSON API data
     include ActionController::MimeResponds
-    before_action :current_user
```
phyrog: Does not really do anything.

spec/graphql/mutations/create_organization_mutation_spec.rb
phyrog: Checks if the hash in the array includes the fields, as include can take other matchers (in this case again include) as argument: https://relishapp.com/rspec/rspec-expectations/v/3-6/docs/built-in-matchers/include-matcher

spec/support/schema.graphql
phyrog: I changed the `Boolean!` to `Boolean`, as this is easier to handle validation and authorization errors (just return null if there are errors). Otherwise these are the same.

lib/tasks/graphql.rake
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+namespace :graphql do
+  desc 'Outputs the current schema'
+  task write: :environment do
+    generated = GraphQL::Schema::Printer.print_schema(OntohubBackendSchema)
+    File.open(Rails.root.join('spec/support/schema.graphql'), 'w') do |f|
+      f.write(generated)
+    end
```
phyrog: The newline would make problems when comparing it to the current schema (see `ontohub_backend_schema_spec.rb`, though, wouldn't it?

app/graphql/types/mutation_type.rb
```diff
@@ -0,0 +1,38 @@
...
+    resolve Mutations::SaveOrganizationMutation.new
+  end
+
+  field :saveAccount, Types::UserType do
+    argument :data, !Types::Input::UserChangesetType
+    argument :password, !types.String
+    resource ->(_obj, _args, ctx) { ctx[:current_user] }
+    resolve Mutations::SaveAccountMutation.new
+  end
+end
```
phyrog: I would just go with alphabetically, since models might not always be clear or multiple models can be affected.

app/graphql/mutations/create_organization_mutation.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+module Mutations
+  # GraphQL mutation to create a new organization and
+  # add the current user as a member
+  class CreateOrganizationMutation
+    def call(_obj, args, ctx)
```
phyrog: Changing it would be fine by me, I'm probably currently more used to JS conventions than Ruby conventions :)

spec/graphql/mutations/create_organization_mutation_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+
+RSpec.describe 'createOrganization mutation' do
+  let!(:current_user) { create :user }
+  let(:organization_data) do
+    {
+      'name' => 'foobar',
+      'displayName' => 'Foobar',
+      'description' => 'This is the foobar',
+    }
+  end
```
phyrog: Uhh, nice!

app/controllers/v2/application_controller.rb
eugenk: I thought it tries to sign in the user?

spec/graphql/mutations/create_organization_mutation_spec.rb
eugenk: Ah ok. That looks good. 

Oh, and please move the line break to the `.to` call:
```ruby
expect(subject['errors']).
  to include(include('message' => 'name is already taken'))
```

lib/tasks/graphql.rake
eugenk: Not if you check it like this:
```ruby
RSpec.describe OntohubBackendSchema do
  let(:current_schema) do
    GraphQL::Schema::Printer.print_schema(OntohubBackendSchema)
  end

  let(:proper_schema) do
    File.read(Rails.root.join('spec/support/schema.graphql')).strip
  end

  it 'matches the generated schema' do
    expect(current_schema).to eq(proper_schema)
  end
end
```

app/graphql/types/mutation_type.rb
eugenk: Alright :+1:

app/graphql/mutations/create_organization_mutation.rb
eugenk: Well, if it's fine with you, let's change it.

app/controllers/v2/application_controller.rb
phyrog: Yeah, but it does not do anything with it if it succeeds and does not do anything if it fails. Before, we used `authenticate_user!` in this spot, which raises an error if a user was not logged in, which we didn't want. Not sure why we didn't just remove the line then.

eugenk: Right. We will do this with pundit anyway...

app/graphql/mutations/create_organization_mutation.rb
phyrog: Would you prefer if I name the unused parameters `_foobar` or just `_`?

eugenk: The convention is `_foobar` because there might be many of them in general.

spec/graphql/mutations/create_organization_mutation_spec.rb
phyrog: Problem is that `values` does not include `slug`

eugenk: Oh yes, because the slug is generated in a before_create hook. You could call `set_slug` on the `organization` object before accessing the `values`.

app/graphql/mutations/save_account_mutation.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+module Mutations
+  # GraphQL mutation to update the current user account
+  class SaveAccountMutation
+    def call(user, args, _ctx)
```
phyrog: It seems easier to me to replicate Devise's update functionality than working around all the things the controller needlessly does (need to create a request and response object, handle the render call, ...).

[update](https://github.com/plataformatec/devise/blob/master/app/controllers/devise/registrations_controller.rb#L44-L63) doesn't look like that much, when stripping away the stuff we won't need (flash, respond_with, bypass_signin, ...).

phyrog: I think stripping all that away, we're left with lines 45 and 48 (and possibly the call to `set_minimum_password_length`, whatever that does), which is the same the mutation currently does.

eugenk: Well, yes, it's simple to recreate it now, but it creates a target for incompatibilities for every devise update.

I think it's okay to do what you suggest, but then please add a comment about it in this mutation and in the Gemfile at the devise-line, so that we can check again when we update devise.

eugenk: `set_minimum_password_length` sets a parameter for the views. It's not important for an API-only backend.

spec/graphql/mutations/save_account_mutation_spec.rb
phyrog: Done in 5271506

app/graphql/types/input/new_organization_type.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+Types::Input::NewOrganizationType = GraphQL::InputObjectType.define do
+  name 'NewOrganization'
+  description <<~DESCRIPTION
+    Contains all fields that are possible to set when creating a new
```
phyrog: Interesting. Seems like Hound does not know the `<<~` syntax?

app/graphql/types/input/new_organization_type.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+Types::Input::NewOrganizationType = GraphQL::InputObjectType.define do
+  name 'NewOrganization'
+
+  argument :name, !types.ID, 'Name of the new organization'
+  argument :displayName, types.String, nil, as: :display_name
+  argument :description, types.String
+end
```
phyrog: I added three test cases to check if all types, fields and arguments have descriptions.

app/graphql/types/mutation_type.rb
```diff
@@ -0,0 +1,80 @@
+# frozen_string_literal: true
+
+# rubocop:disable Metrics/BlockLength
```
phyrog: Well, that didn't work

app/graphql/types/input/new_organization_type.rb
eugenk: It was introduced with Ruby 2.3 and we have 2.4 as the target version in the rubocop config... Looks like Hound runs on 2.2 or less. In any case, let's try to update the rubocop config to 2.4. I've seen a lot of weird behaviour - maybe this changes something...

app/graphql/types/mutation_type.rb
eugenk: That could be a malfunction because of the "syntax error".

Gemfile
```diff
@@ -14,7 +14,7 @@ gem 'puma', '~> 3.0'
 gem 'rack-cors'
 
 gem 'ontohub-models', github: 'ontohub/ontohub-models',
-                      branch: 'master'
+                      branch: '80-organization_membership_is_not_destroyed'
```
eugenk: This should be `master` again.

app/graphql/types/input/new_organization_type.rb
phyrog: Not sure I follow.
> we have 2.4 as the target version

> update the rubocop config to 2.4

eugenk: Sorry, typo. I meant we have 2.3 as the target version

phyrog:
Alright, I think I got everything.

eugenk:
Except for a comment in the Gemfile above the devise-line ;)

phyrog:
We don't have a devise-line in the Gemfile :confused: If you mean the comment about the devise update stuff, I added it to the `save_account_mutation.rb` file.

eugenk:
Oh, yes, devise is in the models and not in the backend. My bad. Maybe we should then add a short comment referring to the mutation above the models-line to keep track of this whenever we update devise in the models.

phyrog:
So, a comment that refers to the other comment? :smile:

phyrog:
Phew. :relieved:

PR #143, opened 2017-07-06
135 Add GraphQL mutations/queries for sessions/signup

### LINKED ISSUE #135: Add GraphQL mutations/queries for sessions/signup
## Types

```graphql
input NewUser {
  username: String!
  displayName: String
  password: String!
  email: String!
}

type SessionToken {
  jwt: String!
  me: User!
}
```

## Queries

```graphql
me: User
```

## Mutations

```graphql
signIn(username: String!, password: String!): SessionToken
signUp(data: NewUser!, captcha: String!): SessionToken

confirmEmail(token: ID!): SessionToken
resendConfirmationEmail(email: String!): Boolean!

resetPassword(token: ID!): SessionToken
resendPasswordResetEmail(email: String!): Boolean!

unlockAccount(token: ID!): SessionToken
resendAccountUnlockEmail(email: String!): Boolean!
```

eugenk:
This shall fix #135 along with some minor code issues (style of RSpec config, consistent use of `ModelURLPath`).

It replicates Devise's code like #136 and actually makes use of Devise internals. This is commented at the appropriate places and we need to remember this when we update Devise. We possibly need to adjust that code on a Devise update.

---

I suggest to review this pull request **commit by commit**


app/graphql/mutations/resend_confirmation_email_mutation.rb
```diff
@@ -0,0 +1,26 @@
...
+    # which would normally be called, performs various things we don't need for
+    # an API only application (including setting flash messages, rendering the
+    # response, bypassing the signin). This code mirrors what is left of the
+    # that action after stripping the rest away.
+    # Make sure that this code is updated if Devise adds relevant code to the
+    # controller action.
+    #
+    # For reference, see the
+    # V2::Users::ConfirmationController#resend_confirmation_email and
+    # https://github.com/plataformatec/devise/blob/7a44233fb9439e7cc4d1503b14f02a1d9f6da7b9/app/controllers/devise/confirmations_controller.rb#L8-L17
```
eugenk: Check again. I changed it by commit-amend.

eugenk: Check again. I changed it by commit-amend.

eugenk: Check again. I changed it by commit-amend.

eugenk: Check again. I changed it by commit-amend.

app/graphql/types/mutation_type.rb
```diff
@@ -45,6 +55,50 @@
...
+    description 'Resends the unlock account email to a user'
+
+    argument :email, !types.String do
+      description 'The email address of the user'
+    end
+
+    resolve Mutations::ResendUnlockAccountEmailMutation.new
+  end
+
+  field :resetPassword, !Types::SessionTokenType do
```
phyrog: I belive this must be of type `Types::SessionTokenType` (without the bang). Background for this is that if an error is set on a field, this field will be set to `null`. If this happens on a field that can't be `null`, the parent field will be set to `null` and so on. Mutations on the other hand are always top-level fields, so if an mutation can raise an error, we must always omit the `!`. Otherwise we would get another error reporting that the field can't be `null`.

eugenk: Fixed by commit-amend

app/graphql/mutations/sign_in_mutation.rb
```diff
@@ -0,0 +1,47 @@
...
+    # Make sure that this code is updated if Devise adds relevant code to the
+    # controller action.
+    #
+    # For reference, see the
+    # V2::Users::SessionsController#create and
+    # https://github.com/plataformatec/devise/blob/7a44233fb9439e7cc4d1503b14f02a1d9f6da7b9/app/controllers/devise/sessions_controller.rb#L16-L22
+    def call(_root, arguments, context)
+      setup_devise(context)
+      transform_params_for_devise(arguments)
+      user = authenticate
```
eugenk: No, because this does the actual authentication for signing in. We don't have the `current_user` at this point.  It is generated inside this method.

PR #151, opened 2017-07-18
Enable fast_finish on travis

phyrog:
This makes Travis not wait for builds that are allowed to fail before reporting the build status to Github, which should improve our feedback duration [bigly](http://memeshappen.com/media/created/2016/12/-Thats-bigly-yuuuge.jpg).


phyrog:
To test this, you can just restart the build on Travis for this branch

PR #152, opened 2017-07-19
Add graphql multiplexing support

phyrog:
Adds support for multiple queries per request; also updates Gems.


spec/controllers/graphql_controller_spec.rb
```diff
@@ -0,0 +1,72 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe GraphqlController do
+  before do
+    # Make rails load the Version model
+    # rubocop:disable Lint/Void
+    Version
+    # rubocop:enable Lint/Void
```
eugenk: I think you can put the rubocop comment in the `Version` line and then, you don't need the `enable` comment:
```ruby
Version # rubocop:disable Lint/Void
```

PR #156, opened 2017-07-22
150 organize graphql

### LINKED ISSUE #150: Find a better way to organize GraphQL types
Putting all types in the top level `types` directory is/will become quite messy, same goes for mutations.

One way we could improve this is grouping related files in subdirectories while keeping the one-type-per-file rule:

```
types/repositories/repository_type.rb
types/repositories/repository_visibility_enum.rb
...
```

Alternatively we could group related types in one file:

```ruby
# types/repositories.rb
module Types
  module Repositories
    RepositoryType = GraphQL::ObjectType.define do
      ...
    end

    RepositoryVisibilityEnum = GraphQL::EnumType.define do
      ...
    end
    ...
  end
end
```

phyrog:
Closes #150.

The structure is now as follows:

- `app/graphql/types/` contains the main types like `Repository`, `Organization` and so on
- `app/graphql/types/organization/` contains types related to organizations (e.g. `OrganizationChangeset`)
- `app/graphql/mutations/account/` contains mutations related to a user account (including sessions). The files now define the fields directly and are only bound to their names in `app/graphql/types/mutation_type.rb`, which means that all code related to one mutation field is in one place
- `spec/graphql/mutations/` mirrors the structure of `app/graphql/mutations`

While this PR looks big, most of it is moving files + moving some code around (which Github does not display very well, unfortunately). It's probably best to review this commit by commit.


app/graphql/mutations/organization/delete_organization_mutation.rb
```diff
@@ -0,0 +1,30 @@
+# frozen_string_literal: true
+
+module Mutations
+  module Organization
+    DeleteOrganizationMutation = GraphQL::Field.define do
+      type types.Boolean
+      description <<~DESCRIPTION
```
phyrog: Ignore

app/graphql/mutations/organization/delete_organization_mutation.rb
```diff
@@ -0,0 +1,30 @@
...
+        Deletes an organization.
```
phyrog: Ignore

app/graphql/mutations/organization/delete_organization_mutation.rb
```diff
@@ -0,0 +1,30 @@
...
+        Returns `true` if it was successful and `null` if there was an error.
```
phyrog: Ignore

app/graphql/mutations/organization/delete_organization_mutation.rb
```diff
@@ -0,0 +1,30 @@
...
+      DESCRIPTION
```
phyrog: Ignore

app/graphql/mutations/account/delete_account_mutation.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+module Mutations
+  module Account
+    DeleteAccountMutation = GraphQL::Field.define do
+      type types.Boolean
+      description <<~DESCRIPTION
+        Deletes the account of the currently signed in user.
```
phyrog: Ignore

PR #162, opened 2017-07-25
Split validation errors

phyrog:
At the moment, validation errors will be concatenated to one long string; this PR returns each validation error as a single `errors` entry, which makes it easier to display them in the frontend.


phyrog:
Heh, I almost used that gif in #161.

PR #164, opened 2017-07-28
Create search graphql types and queries

### LINKED ISSUE #138: Add graphql query for search
## Types

```graphql
union SearchResultTarget = Repository | Organization | User

type SearchResultEntry {
  ranking: Float!
  entry: SearchResultTarget!
}

type SearchResultCount {
  # should include all possible values of SearchCategory as keys
  all: Int!
  organizationalUnits: Int!
  repositories: Int!
}

type SearchResult {
  entries: [SearchResultEntry!]!
  count: SearchResultCount!
}

enum SearchCategory {
  all
  organizationalUnits
  repositories
}
```

## Queries

```graphql
search(query: String!, category: SearchCategory!, limit: Int, skip: Int): SearchResult!
```

If we decide not to send the Elasticsearch ranking with each result, we can simplify `SearchResultEntry` to `SearchResultTarget`.

phyrog:
Closes #138.


spec/graphql/queries/search_query_spec.rb
```diff
@@ -0,0 +1,143 @@
...
+    end
+
+    it 'returns the organizational units' do
+      organizational_units = search_result['entries'].select do |e|
+        %w(User Organization).include?(e['entry']['__typename'])
+      end
+      expect(organizational_units.length).to eq(5)
+    end
+  end
+end
```
eugenk: This is only a comment, not a change request:

The most part of the specs is duplicated code. You could write a shared example block that uses let-defined variables that you check on, as in:
```ruby
    it 'returns the correct number of entries' do
      expect(search_result['entries'].length).to eq(expected_num_entries)
      expect(search_result['count']['all']).to eq(expected_count_all)
      expect(search_result['count']['organizationalUnits']).to eq(expected_count_organizational_units)
      expect(search_result['count']['repositories']).to eq(expected_count_repositories)
    end
```
This would make the test clearer.

app/graphql/types/query_type.rb
```diff
@@ -38,6 +38,25 @@
...
+      default_value 'all'
+    end
+
+    resolve(lambda do |_root, arguments, _context|
+      category = arguments['category'] ||
+        target.arguments['category'].default_value
+      SearchResult.new(nil).to_struct(category: category)
+    end)
+  end
+
```
phyrog: So, queries like this?

```graphql
{
  search(query: "foobar") {
    repository(id: "ada/repo0", category: symbols) {
      count {
        ...
      }
      results {
        ...
      }
    }
    site(category: repositories) {
      count {
        ...
      }
      results {
        ...
      }
    }
  }
}
```

I think I like it.

app/graphql/types/query_type.rb
```diff
@@ -38,6 +38,25 @@
...
   end
 
+  field :search, !Types::SearchResultType do
+    description 'Search Ontohub'
+
+    argument :query, !types.String do
+      description 'The query string'
+    end
+
+    argument :category, Types::SearchResult::CategoryType do
```
phyrog: I was orientating on the [GitHub search](https://github.com/search?q=foobar), where you can see the number of results for all categories, but only see one category of results at a time. That's also the reason for the structure with the `entries` and `count` fields.

That means that when you switch to a different tab, you do a new search request with a different category argument. I'm open for other ideas however, as there are certainly downsides to this.

eugenk: Ah, okay. So filtering for the categories already happens on the server. This would definitely save some bandwidth and probably save some time during the search. It's good to have it restricted like this for the frontend, but API users might want to be able to query for multiple categories at once.

I suggest:
Allow to query for multiple categories at once in the API, but only specify one in a usual search query in the frontend. This also allows us to select multiple categories in the future if users give us such feedback.

app/graphql/types/query_type.rb
eugenk: Yes, like this. Looks good.

app/models/search_result.rb
```diff
@@ -23,4 +23,40 @@ def initialize(_params)
...
+      OpenStruct.new(entry: RepositoryCompound.wrap(repo), ranking: 1.0)
+    end
+    org_units = organizational_units.map do |org_unit|
+      OpenStruct.new(entry: org_unit, ranking: 1.0)
+    end
+
+    result.entries = []
+    if !params[:categories] ||
+       params[:categories].empty? ||
+       params[:categories].include?('repositories')
```
eugenk: I think this can be made simpler:
```ruby
if params[:categories].blank? || params[:categories].include?('repositories')
```
See http://api.rubyonrails.org/classes/Object.html#method-i-blank-3F

app/models/search_result.rb
```diff
@@ -23,4 +23,40 @@ def initialize(_params)
...
+      result.entries.concat(repos)
+    end
+    if !params[:categories] ||
+       params[:categories].empty? ||
+       params[:categories].include?('organizationalUnits')
```
eugenk: same here

app/graphql/types/search_result/global_scope/target_type.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+Types::SearchResult::GlobalScope::TargetType = GraphQL::UnionType.define do
```
eugenk: I'm not sure if `target` is the best choice for the name. When I read `target`, I think of what we call `category`. How about `SpecificEntry`?

It's also weird to have the field `global.entries.entry` to be a `GlobalSearchTarget` instead of something with "entry" in it ;)

phyrog: Well I mean, the `target` types kind of mirror the `categories`. I chose `target` because  the `entry` points to, well, the target :)

I think a type named `SpecificGlobalSearchEntry` might be confusing, since we already have the `GlobalSearchEntry`, so maybe something else entirely?

app/models/search_result.rb
phyrog: Fixed in 2664dac.

app/models/search_result.rb
phyrog: Also fixed in 2664dac

app/graphql/types/search_result/global_scope/target_type.rb
eugenk: What about this?
GlobalSearchEntry -> RankedGlobalSearchEntry
GlobalSearchTarget -> GlobalSearchEntry

phyrog: Sounds good to me

PR #170, opened 2017-07-31
Add role filter argument to user's organizations field

### LINKED ISSUE #169: Return a users administrable namespaces
Needed to figure out, in which namespaces a user can create a repository. A list of `String` or a list of actual `OrganizationalUnit`s would both be fine.

phyrog:
Closes #169. I decided to just use the `organizations` field and add a `role` filter argument.

### DIFF
--- a/app/graphql/types/user_type.rb
+++ b/app/graphql/types/user_type.rb
@@ -34,8 +34,17 @@ Types::UserType = GraphQL::ObjectType.define do
       default_value 0
     end
 
+    argument :role, Types::Organization::RoleEnum do
+      description 'Filter the organizations by the user\'s role'
+    end
+
     resolve(lambda do |user, arguments, _context|
-      user.organizations_dataset.order(:slug).
+      dataset = if arguments['role']
+                  user.organizations_by_role(arguments['role'])
+                else
+                  user.organizations_dataset
+                end
+      dataset.order(:slug).
         limit(arguments['limit'], arguments['skip'])
     end)
   end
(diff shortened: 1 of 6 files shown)


app/graphql/types/user_type.rb
```diff
@@ -34,8 +34,17 @@
       default_value 0
     end
 
+    argument :role, Types::Organization::RoleEnum do
+      description 'Filter the organizations by the user\'s role'
```
eugenk: Please use a double quoted String to avoid escaping.

spec/graphql/types/user_type_spec.rb
```diff
@@ -44,5 +48,27 @@
...
+      it 'returns the organizations with the role: admin' do
+        role = 'admin'
+        organizations = organizations_field.resolve(
+          user,
+          organizations_field.default_arguments('role' => role),
+          {}
+        )
+        expect(organizations.count).to eq(1)
+      end
+    end
```
eugenk: Why not
```ruby
    let(:organizations) do
      organizations_field.
        resolve(user,
                organizations_field.default_arguments('role' => role),
                {})
    end
    context 'filters by role' do
      context 'read' do
        let(:role) { 'read' }
        it 'returns the organizations with the corresponding role' do
          expect(organizations.count).to eq(20)
        end
      end

      context 'admin' do
        let(:role) { 'admin' }
        it 'returns the organizations with the corresponding role' do
          expect(organizations.count).to eq(1)
        end
      end
    end
```
I find this much easier to read and maintain because you DRY the complicated stuff.

phyrog:
Did you just spoiler yourself?

eugenk:
I did not look too much at it (only the tiny version) so I don't know what's going on there. What I saw was just a random handshake :)

PR #172, opened 2017-08-03
171 Fix errors in Git for GraphQL

### LINKED ISSUE #171: Fix errors in Git for GraphQL
# Directory
```
query {
  repository(id:"ada/repo0") {
    commit(revision:"master") {
      directory(path:"inexistant") {
        path
      }
    }
  }
}
```
returns
```
{
  "data": {
    "repository": {
      "commit": null
    }
  },
  "errors": [
    {
      "message": "Cannot return null for non-nullable field Commit.directory"
    }
  ]
}
```

-----
# File
```
query {
  repository(id:"ada/repo0") {
    commit(revision:"master") {
      file(path:"inexistant") {
        path
      }
    }
  }
}
```
fails with
```
NoMethodError (undefined method `name' for nil:NilClass):

app/models/git_file.rb:16:in `initialize'
app/graphql/types/git/commit_type.rb:97:in `new'
app/graphql/types/git/commit_type.rb:97:in `block (3 levels) in <top (required)>'
app/graphql/instrumenters/validation_error_instrumenter.rb:16:in `block (2 levels) in instrument'
app/controllers/graphql_controller.rb:10:in `execute'
```

eugenk:
This shall fix #171. It also upgrades gitlab_git and adds a few more fields.

### DIFF
--- a/app/graphql/mutations/repository/git/set_default_branch_mutation.rb
+++ b/app/graphql/mutations/repository/git/set_default_branch_mutation.rb
@@ -1,0 +1,41 @@
+# frozen_string_literal: true
+
+module Mutations
+  module Repository
+    module Git
+      SetDefaultBranchMutation = GraphQL::Field.define do
+        type Types::Git::BranchType
+        description 'Sets the default branch'
+
+        argument :repositoryId, !types.ID do
+          description 'The repository to create the tag in'
+        end
+
+        argument :name, !types.String do
+          description 'The name of the branch to make it the default'
+        end
+
+        resource(lambda do |_root, arguments, _context|
+          RepositoryCompound.find(slug: arguments['repositoryId'])
+        end)
+
+        resolve SetDefaultBranchResolver.new
+      end
+
+      # GraphQL mutation to create a new tag
+      class SetDefaultBranchResolver
+        def call(repository, arguments, context)
+          arguments = arguments.to_h
+          if repository.git.find_branch(arguments['name'])
+            repository.git.default_branch = arguments['name']
+            repository.git.find_branch(repository.git.default_branch)
+          else
+            message = %(The branch "#{arguments['name']}" does not exist.)
+            context.add_error(GraphQL::ExecutionError.new(message))
+            nil
+          end
+        end
+      end
+    end
+  end
+end
--- a/app/graphql/types/git/commit_type.rb
+++ b/app/graphql/types/git/commit_type.rb
@@ -101,6 +101,15 @@ Types::Git::CommitType = GraphQL::ObjectType.define do
     end)
   end
 
+  field :lsFiles, !types[!types.String] do
+    description 'A list of all file paths in the repository'
+
+    resolve(lambda do |commit, _arguments, _context|
+      gitlab_wrapper = Gitlab::Git::Wrapper.new(commit.repository.path)
+      gitlab_wrapper.ls_files(commit.id)
+    end)
+  end
+
   field :diff, !types[Types::Git::DiffType] do
     description 'The changes that this commit introduced'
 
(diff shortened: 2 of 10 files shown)


app/graphql/mutations/repository/git/set_default_branch_mutation.rb
```diff
@@ -0,0 +1,41 @@
...
+
+      # GraphQL mutation to create a new tag
+      class SetDefaultBranchResolver
+        def call(repository, arguments, context)
+          arguments = arguments.to_h
+          if repository.git.find_branch(arguments['name'])
+            repository.git.default_branch = arguments['name']
+            repository.git.find_branch(repository.git.default_branch)
+          else
+            message = %(The branch "#{arguments['name']}" does not exist.)
```
phyrog: Why this form of string interpolation?

app/graphql/types/git/commit_type.rb
```diff
@@ -94,9 +94,19 @@
...
-                  arguments['path'],
-                  load_all_data: arguments['loadAllData'])
+      git_file = GitFile.new(commit,
+                             arguments['path'],
+                             load_all_data: arguments['loadAllData'])
+      git_file if git_file.exist?
+    end)
+  end
+
+  field :lsFiles, !types[!types.String] do
```
phyrog: What do you think about returning a list of files here instead of just a list of paths? Would that be computationally expensive to do?

app/models/git_file.rb
```diff
@@ -11,35 +11,55 @@ class GitFile
...
-    blob.binary ? 'base64' : 'plain'
+    blob&.binary ? 'base64' : 'plain'
   end
 
   def gitlab
     @gitlab ||= Gitlab::Git::Wrapper.new(@commit.repository.path)
   end
 
+  def exist?
+    unless @existance_checked
```
phyrog: I think you might write this as

```ruby
@existance_checked ||= !!check_existance
```

Not sure if it's more readable though

app/models/git_file.rb
```diff
@@ -11,35 +11,55 @@ class GitFile
...
+  # it, we can load the path as well.
+  def load_name_and_path_from_blob
+    check_existance
+    return unless blob
+    @name = blob.name
+    @path = blob.path
+  end
+
+  # Checks the existance by loading the blob
+  def check_existance
```
phyrog: Shouldn't this set `@exiatance_checked`?

app/graphql/mutations/repository/git/set_default_branch_mutation.rb
eugenk: Because I want to use double quotes in the string and don't want to escape them.

phyrog: Ah ok, now I get it

app/graphql/types/git/commit_type.rb
eugenk: We could do it without blowing up the computation time if we only query for `name` and `path`. However, it would be a blob lookup (including loading up to 10 MB of the data into memory per blob) for every file of the repository (some ontology repositories have many files) as soon as we query for more than just path and name.

I vote for not doing this.

app/models/git_file.rb
eugenk: Simplified in fdc916bb

app/models/git_file.rb
eugenk: Simplified in fdc916bb

app/models/git_file.rb
phyrog: Ah wait that is not the same (if `check_existance` returns a falsy value)

eugenk:
Rebased on master

PR #182, opened 2017-08-11
Create GraphQl-REST controller for repositories

### LINKED ISSUE #158: Create GraphQL-REST controller for repositories
We need the following routes:

```
GET /:organizational_unit_id/repositories => `v3/repositories_controller#index`
GET /:organizational_unit_id/:repository_id => `v3/repositories_controller#show`
```

These should use the GraphQL API, as can be seen in `app/controllers/v3/version_controller.rb` and `app/controllers/v3/organizational_units_controller.rb`.

## `show`

should return these repository fields:
- `branches`
- `content_type`
- `default_branch`
- `description`
- `id`
- `name`
- `owner`: `id`
- `visibility`

## `index`

almost the same fields as `show`, only as a list of repositories owned by the organizational_unit, with this difference:
- `owner`: `id` should be left out and instead be returned by the parent `organizationalUnit` field

DerProfessor:
Closes #158

### DIFF
--- a/spec/controllers/v3/repositories_controller_spec.rb
+++ b/spec/controllers/v3/repositories_controller_spec.rb
@@ -17,9 +17,9 @@ RSpec.describe V3::RepositoriesController do
     end
 
     describe 'action: show' do
-      before do 
-        get :show, 
-          params: {organizational_unit_id: user.to_param, 
+      before do
+        get :show,
+          params: {organizational_unit_id: user.to_param,
                    repository_id: subject.to_param.sub(%r{\A[^/]*/}, '')}
       end
 
@@ -41,7 +41,7 @@ RSpec.describe V3::RepositoriesController do
 
     describe 'action: show' do
       before do
-        get :index, params: {organizational_unit_id: '', repository_id:''}
+        get :index, params: {organizational_unit_id: '', repository_id: ''}
       end
       it 'Does not find the repository.' do
         expect(response_data['repository']).to be(nil)
--- a/spec/controllers/v3/repositories_controller_spec.rb
+++ b/spec/controllers/v3/repositories_controller_spec.rb
@@ -1,0 +1,52 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe V3::RepositoriesController do
+  subject { create :repository_compound }
+  let!(:user) { subject.owner }
+
+  context 'successful' do
+    describe 'action: index' do
+      before do
+        get :index, params: {organizational_unit_id: user.to_param}
+      end
+
+      it { expect(response).to have_http_status(:ok) }
+      it { |example| expect([example, response]).to comply_with_rest_api }
+    end
+
+    describe 'action: show' do
+      before do 
+        get :show, 
+          params: {organizational_unit_id: user.to_param, 
+                   repository_id: subject.to_param.sub(%r{\A[^/]*/}, '')}
+      end
+
+      it { expect(response).to have_http_status(:ok) }
+      it { |example| expect([example, response]).to comply_with_rest_api }
+    end
+  end
+
+  context 'failing because of bad params' do
+    describe 'action: index' do
+      before do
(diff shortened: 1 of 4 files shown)


spec/controllers/v3/repositories_controller_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+
+  context 'failing because of bad params' do
+    describe 'action: index' do
+      before do
+        get :index, params: {organizational_unit_id: ''}
+      end
+      it 'Does not find the organizational unit.' do
+        expect(response_data['organizationalUnit']).to be(nil)
+      end
+      it { expect(response).to have_http_status(:ok) }
```
phyrog: Please add another check, that the returned data still complies with the api schema. You can just use line you used above:
```ruby
it { |example| expect([example, response]).to comply_with_rest_api }
```

spec/controllers/v3/repositories_controller_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+    end
+
+    describe 'action: show' do
+      before do
+        get :index, params: {organizational_unit_id: '', repository_id: ''}
+      end
+      it 'Does not find the repository.' do
+        expect(response_data['repository']).to be(nil)
+      end
+      it { expect(response).to have_http_status(:ok) }
```
phyrog: Same as above, please add another `complies_with_rest_api` check.

spec/controllers/v3/repositories_controller_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+  subject { create :repository_compound }
+  let!(:user) { subject.owner }
+
+  context 'successful' do
+    describe 'action: index' do
+      before do
+        get :index, params: {organizational_unit_id: user.to_param}
+      end
+
+      it { expect(response).to have_http_status(:ok) }
```
phyrog: Please add another `it` that checks if `repository` and `organizationalUnit` is not nil.
```ruby
it 'finds the organizational unit' do
  expect(response_data['organizationalUnit']).not_to be(nil)
end
it 'finds the repository' do
  expect(response_data['organizationalUnit']['repository']).not_to be(nil)
end
```

spec/controllers/v3/repositories_controller_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+
+    describe 'action: show' do
+      before do
+        get :show,
+          params: {organizational_unit_id: user.to_param,
+                   repository_id: subject.to_param.sub(%r{\A[^/]*/}, '')}
+      end
+
+      it { expect(response).to have_http_status(:ok) }
+      it { |example| expect([example, response]).to comply_with_rest_api }
```
phyrog: Please add another `it` that checks if `repository` is not nil.
```ruby
it 'finds the repository' do
  expect(response_data['repository']).not_to be(nil)
end
```

spec/controllers/v3/repositories_controller_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+    end
+  end
+
+  context 'failing because of bad params' do
+    describe 'action: index' do
+      before do
+        get :index, params: {organizational_unit_id: ''}
+      end
+      it 'Does not find the organizational unit.' do
```
phyrog: Nitpick: I think all other `it`s start with lowercase and don't end with `.`:

```ruby
it 'does not find the organizational unit' do
```

spec/controllers/v3/repositories_controller_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+        expect(response_data['organizationalUnit']).to be(nil)
+      end
+      it { expect(response).to have_http_status(:ok) }
+    end
+
+    describe 'action: show' do
+      before do
+        get :index, params: {organizational_unit_id: '', repository_id: ''}
+      end
+      it 'Does not find the repository.' do
```
phyrog: See other nitpick

PR #183, opened 2017-08-11
167 set up asynchronous computing

### LINKED ISSUE #167: Set up asynchronous computing
Some tasks like sending emails or evaluate an OMS need to be performed asynchronously (in the background) without blocking the HTTP response.

Set up [ActiveJob](http://edgeguides.rubyonrails.org/active_job_basics.html) with the [backend](http://edgeapi.rubyonrails.org/classes/ActiveJob/QueueAdapters.html) "sneakers". This requires RabbitMQ, which we settled upon as our message broker.

Setup devise to [send the user-account-specific emails asynchronously](https://github.com/plataformatec/devise/wiki/How-To:-Send-devise-emails-in-background-(Resque,-Sidekiq-and-Delayed::Job)).

Furthermore, change the wiki [setup page](https://github.com/ontohub/ontohub-backend/wiki/Setting-up-the-development-environment) to reflect the configuration of the overall development setup, especially RabbitMQ (if needed at all).

phyrog:
Closes #167.


phyrog:
@eugenk, can you check out the WIP commit? I'm having trouble with the `resend_unlock_account_email_mutation_spec`. I'm doing the same thing as in the other mutation specs and controller specs (even the unlock controller spec works), but for some reason the email job is just not queued. You don't need rabbitmq for this, tests are running the `ActiveJob::QueueAdapters::TestAdapter` instead of sneakers.

phyrog:
I think I figured it out, but it seems odd to me. The unlocking mail seems to be sent by the `user.lock_access!` call, not by the graphql query itself. Could it be that the test is incorrect here?

eugenk:
> The unlocking mail seems to be sent by the user.lock_access! call, not by the graphql query itself. 

Correct. All emails are sent directly through model methods, not mutation/controller methods. In the API we don't care about these (sometimes private) model methods. We only want to be sure that side effects like sending emails are performed.

> Could it be that the test is incorrect here?

What test?

phyrog:
There are actually two emails sent during the test run. One is sent by the explicit locking of the user (`user.lock_access!`) and one is sent by the `ResendUnlockAccountEmail` mutation. I think before the latest commit, the test actually checked the first email (sent by the locking of the user) instead of the second email (which I don't think was ever sent, because we were calling `subject` (and thus the mutation) in the before hook before we actually locked the user, which would result in no email being sent.

I believe the latest commit fixes this

spec/controllers/v2/users/passwords_controller_spec.rb
```diff
@@ -37,10 +40,12 @@
     context 'successful' do
       before do
         token = user.send_reset_password_instructions
-        UsersMailer.deliveries.clear
```
eugenk: Why this deletion?

spec/controllers/v2/users/confirmations_controller_spec.rb
```diff
@@ -41,9 +41,11 @@
   describe 'POST resend_confirmation_email',
     type: :mailer, no_transaction: true do
     before do
-      UsersMailer.deliveries.clear
```
eugenk: Why this deletion?

spec/controllers/v2/users/account_controller_spec.rb
```diff
@@ -86,7 +92,7 @@
         end
 
         it 'sends a confirmation email and two notification emails' do
-          expect(UsersMailer.deliveries.size).to eq(3)
```
eugenk: Please don't delete this check. I'd still like to know if all emails have actually been delivered.

spec/controllers/v2/users/sessions_controller_spec.rb
```diff
@@ -66,32 +66,38 @@
...
+                format: :json
+            end
           end
 
           it 'does not lock the user' do
             expect(user.reload.access_locked?).to be(false)
           end
 
           it 'does not send an email' do
-            expect(UsersMailer.deliveries).to be_empty
```
eugenk: Please don't delete this check. I'd still like to know if all emails have actually been delivered.

spec/controllers/v2/users/sessions_controller_spec.rb
```diff
@@ -100,7 +106,7 @@
 
           context 'unlock instructions email' do
             it 'sends an instructions email' do
-              expect(UsersMailer.deliveries.size).to eq(1)
```
eugenk: Please don't delete this check. I'd still like to know if all emails have actually been delivered.

spec/graphql/mutations/account/resend_confirmation_email_mutation_spec.rb
```diff
@@ -29,7 +31,7 @@
   subject { result }
 
   before do
-    UsersMailer.deliveries.clear
```
eugenk: Why this deletion?

spec/graphql/mutations/account/resend_password_reset_email_mutation_spec.rb
```diff
@@ -29,7 +31,7 @@
   subject { result }
 
   before do
-    UsersMailer.deliveries.clear
```
eugenk: Why this deletion?

spec/graphql/mutations/account/resend_unlock_account_email_mutation_spec.rb
```diff
@@ -28,17 +30,14 @@
 
   subject { result }
 
-  before do
-    UsersMailer.deliveries.clear
-    subject
-  end
-
```
eugenk: Why this deletion?

spec/graphql/mutations/account/resend_unlock_account_email_mutation_spec.rb
```diff
@@ -53,7 +52,7 @@
       end
 
       it 'does not send an email' do
-        expect(UsersMailer.deliveries).to be_empty
```
eugenk: Please don't delete this check. I'd still like to know if all emails have actually been delivered.

spec/graphql/mutations/account/reset_password_mutation_spec.rb
```diff
@@ -37,7 +39,7 @@
   subject { result }
 
   before do
-    UsersMailer.deliveries.clear
```
eugenk: Why this deletion?

spec/shared_examples/confirmation_email.rb
```diff
@@ -2,7 +2,7 @@
 
 RSpec.shared_examples 'a confirmation email sender' do
   it 'sends an email' do
-    expect(UsersMailer.deliveries.size).to eq(1)
```
eugenk: Please don't delete this check. I'd still like to know if all emails have actually been delivered.

spec/shared_examples/password_has_been_reset_email.rb
```diff
@@ -2,7 +2,7 @@
 
 RSpec.shared_examples 'a password has been reset email sender' do
   it 'sends a notification email' do
-    expect(UsersMailer.deliveries.size).to eq(1)
```
eugenk: Please don't delete this check. I'd still like to know if all emails have actually been delivered.

spec/shared_examples/resend_password_reset_email.rb
```diff
@@ -2,7 +2,7 @@
 
 RSpec.shared_examples 'a password reset email sender' do
   it 'sends an instructions email' do
-    expect(UsersMailer.deliveries.size).to eq(1)
```
eugenk: Please don't delete this check. I'd still like to know if all emails have actually been delivered.

spec/shared_examples/resend_unlock_account_email.rb
```diff
@@ -2,7 +2,7 @@
 
 RSpec.shared_examples 'an unlock account email sender' do
   it 'sends an instructions email' do
-    expect(UsersMailer.deliveries.size).to eq(1)
```
eugenk: Please don't delete this check. I'd still like to know if all emails have actually been delivered.

eugenk:
Now that I've seen all the changes, I can answer the question "Why this deletion" myself. However, I would like it to stay in there because there is a difference between performing the job and actually delivering the emails. The latter is the stronger property because there might be something wrong with the mailer (not the worker).

lib/sneakers/workers.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+module Sneakers
+  # Module to contain the created Worker classes
```
phyrog: I didn't find the issue, but I added a comment that this is probably only a temporary solution and we will want to switch to actual worker class definitions.

phyrog: Ah what the hell, I'll do it in this PR, because why wait...

config/initializers/settings_normalizer.rb
```diff
@@ -36,4 +37,14 @@ def normalize_path(path)
 
     Rails.root.join(path)
   end
+
+  def normalize_worker_groups
```
eugenk: Please add a test in spec/initializers/settings_normalizer_spec.rb

config/initializers/settings_normalizer.rb
```diff
@@ -36,4 +37,14 @@ def normalize_path(path)
...
+    groups = @settings.sneakers
+    groups.map do |group|
+      classes = group.classes
+      classes = [classes] unless classes.is_a?(Array)
+      group.classes = classes
```
eugenk: You can simply convert it to an array by replacing the three lines above with
```ruby
group.classes = Array(group.classes)
```

config/settings.yml
```diff
@@ -8,3 +8,7 @@ jwt:
 # interpreted relative to the application root directory.
 # This directory will be created if it does not exist.
 data_directory: 'data'
+
+sneakers:
+  - classes: MailersWorker
+    workers: 1
```
eugenk: Please add validators in config/initializers/settings_validator.rb and specs in spec/initializers/settings_validator_spec.rb

config/initializers/settings_normalizer.rb
phyrog: Yea, but that would be too easy.

Issue #175, opened 2017-08-05
Integrate graphql-pundit


phyrog:
Authorize fields:
- [x] `UserType#email` (#201)
- [x] `UserType#unconfirmedEmail` (#201)
- [x] `UserType#publicKeys` (#201)
- [x] `QueryType#repository` (#201)

Authorize scopes:
- [x] `OrganizationalUnitType#repositories` (#201)
- [ ] `GlobalSearchScope#entries` probably only once search is implemented properly

Authorizes mutations:
- [x] account mutations
  - [x] `AddPublicKeyMutation` (#219)
  - [x] `ConfirmEmailMutation` (#219)
  - [x] `DeleteAccountMutation` (#219)
  - [x] `RemovePublicKeyMutation` (#219)
  - [x] `ResendConfirmationEmailMutation` (#219)
  - [x] `ResendPasswordResetEmailMutation` (#219)
  - [x] `ResendUnlockAccountEmailMutation` (#219)
  - [x] `ResetPasswordMutation` (#219)
  - [x] `SaveAccountMutation` (#219)
  - [x] `SignInMutation` (#219)
  - [x] `SignUpMutation` (#219)
  - [x] `UnlockAccountMutation` (#219)
- [x] organization mutations
  - [x] `CreateOrganizationMutation` (#221)
  - [x] `DeleteOrganizationMutation` (#221)
  - [x] `SaveOrganizationMutation` (#221)
- [x] repository mutations
  - [x] `CreateRepositoryMutation` (#222)
  - [x] `DeleteRepositoryMutation` (#222)
  - [x] `SaveRepositoryMutation` (#222)
  - [x] git mutations
    - [x] `CommitMutation` (#224)
    - [x] `CreateBranchMutation` (#224)
    - [x] `CreateTagMutation` (#224)
    - [x] `DeleteBranchMutation` (#224)
    - [x] `DeleteTagMutation` (#224)
    - [x] `SetDefaultBranchMutation` (#224)

As this requires extensive tests, this can be done in multiple PRs

PR #201, opened 2017-09-06
Integrate graphql-pundit

phyrog:
Part of #175. Authorizes some fields:

- UserType#email, UserType#public_keys, UserType#unconfirmed_email are authorized
- OrganizationalUnitType#repositories is scoped
- QueryType#repository is authorized

Also I added a test for authorizing the REST branches controller


app/policies/user_policy.rb
```diff
@@ -11,6 +11,6 @@ def show_current_user?
   end
 
   def access_private_data?
-    current_user == resource
+    !!current_user && current_user.id == resource.id
```
eugenk: Redundant `!!`:
`&&` will always return `true` or `false`.

phyrog: Hm I had a problem that it returned `nil` instead of `false` when `current_user` was `nil`; will check again.

phyrog:
```ruby
nil && 4 == 5
=> nil
```

eugenk: Ah okay. Sorry for misleading you.

PR #202, opened 2017-09-11
Bundle update.

eugenk:
This upgrades the bundle. Sequel 5.0.0 is out there, but [it is not yet supported by sequel-rails](https://github.com/TalentBox/sequel-rails/issues/142).


greysteil:
@eugenk - would you be up for automatically creating dependency update PRs? I built [Dependabot](https://dependabot.com) to do that at my last company and we made it free for open-source - could be a time saver? I'm trying to get as many people using it as possible so the [compatibility scores](https://dependabot.com/compatibility-score.html) are accurate.

eugenk:
@greysteil, thanks for suggesting this. We'll try dependabot. Seems like it can save us some dull work.

PR #220, opened 2017-10-03
Fix NoMethodError: undefined method join for String

eugenk:
This fixes the following error

```text
NoMethodError (undefined method `join' for "data":String):

app/models/repository_compound.rb:26:in `git_directory'
app/models/repository_compound.rb:62:in `git_path'
app/models/repository_compound.rb:56:in `git'
app/graphql/types/repository_type.rb:95:in `block (3 levels) in <top (required)>'
app/graphql/instrumenters/validation_error_instrumenter.rb:16:in `block (2 levels) in instrument'
app/controllers/graphql_controller.rb:10:in `execute'
```

The problem was that [config adds a `before_action` to reload the settings on every request in the development environment](https://github.com/railsconfig/config/blob/1.4.0/lib/config/integrations/rails/railtie.rb#L26-L32). This does not post-process the `Settings`. This pull request monkey-patches `Config`'s public API to append our post-processing.

[This has already been there about seven years ago](https://github.com/railsconfig/config/blob/53257274be8818f6293cab82d40cf4c3a7dfe8b2/lib/rails_config/integration/rails.rb#L25-L27).

### DIFF
--- a/spec/support/simplecov.rb
+++ b/spec/support/simplecov.rb
@@ -8,6 +8,9 @@ unless defined?(SimpleCov)
 
     # The config of the 'config' gem does not need to be tested.
     add_filter 'config/initializers/config.rb'
+    # The monkey-patch for the development mode of the 'config' gem does not
+    # need to be tested.
+    add_filter 'config/initializers/core_extensions/config.rb'
   end
   require 'codecov'
   formatters = [SimpleCov::Formatter::HTMLFormatter]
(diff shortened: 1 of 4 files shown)


spec/support/simplecov.rb
```diff
@@ -8,6 +8,9 @@
 
     # The config of the 'config' gem does not need to be tested.
     add_filter 'config/initializers/config.rb'
+    # The monkey-patch for the development mode of the 'config' gem does not
+    # need to be tested.
+    add_filter 'onfig/initializers/core_extensions/config.rb'
```
phyrog: Typo here?

eugenk: Yes. Fixed by commit amend.

PR #222, opened 2017-10-04
175 authorize repository mutations

phyrog:
Part of #175.

### DIFF
--- a/spec/graphql/mutations/repository/create_repository_mutation_spec.rb
+++ b/spec/graphql/mutations/repository/create_repository_mutation_spec.rb
@@ -82,4 +82,14 @@ RSpec.describe 'createRepository mutation' do
         to include(include('message' => "You're not authorized to do this"))
     end
   end
+
+  context 'Not authorized for the namespace' do
+    let(:context) { {current_user: create(:user)} }
+
+    it 'returns an error' do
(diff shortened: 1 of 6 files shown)


spec/graphql/mutations/repository/create_repository_mutation_spec.rb
```diff
@@ -72,4 +72,14 @@
...
+
+  context 'Not signed in' do
+    let(:context) { {} }
+
+    it 'returns an error' do
+      expect(subject['data']['createRepository']).to be(nil)
+      expect(subject['errors']).
+        to include(include('message' => "You're not authorized to do this"))
+    end
+  end
```
eugenk: This is the only case where I would like to have another example testing if you are authorised to perform the action if you are signed in as a "bad user" (not the owner). It would be checking if the resource `owner` passed to `authorize!` is the correct one.

https://github.com/ontohub/ontohub-backend/blob/f9f92395b957e908572cf8b893c51972b5f3d9b4/app/graphql/mutations/repository/create_repository_mutation.rb#L13-L19

phyrog: Done in 3432f19

PR #238, opened 2017-10-12
186 create graphql rest controller for git log

### LINKED ISSUE #186: Create GraphQL-REST Controller for Git Log
# Log
## index
* routes
  The `*path` is a wildcard path segment. It allows any character to occur in this segment (also slashes). The router maps it to `params[:path]` in the controller. The controller action is supposed to show the git log of the revision. If no revision is supplied by the user, use the default branch.
  * `/v3/:organizational_unit_id/:repository_id/history/*path`
  * `/v3/:organizational_unit_id/:repository_id/revision/:revision/history/*path`
* fields
  * `id`
  * `message`
  * `committedAt`
* optional arguments (additional to the ones in the URL) - supposed to be supplied in a query string
  * `limit`
  * `skip`
  * `skipMerges`
  * `before`
  * `after`

DerProfessor:
Close #186

### DIFF
--- a/spec/controllers/rest/history_controller_spec.rb
+++ b/spec/controllers/rest/history_controller_spec.rb
@@ -1,0 +1,71 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe Rest::HistoryController do
+  let!(:repository) { create :repository_compound, :not_empty }
+  let!(:user) { repository.owner }
+  let!(:path) { repository.git.ls_files(revision).first }
+  let!(:revision) { repository.git.default_branch }
+
+  context 'successful (without revision)' do
+    describe 'action: index' do
+      before do
+        get :index, params: {organizational_unit_id: user.to_param,
+                              repository_id:
+                              repository.to_param.sub(%r{\A[^/]*/}, ''),
+                              path: path}
+      end
+      it { expect(response).to have_http_status(:ok) }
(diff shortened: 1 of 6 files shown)


spec/controllers/rest/history_controller_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+        get :index, params: {organizational_unit_id: user.to_param,
+                              repository_id:
+                              repository.to_param.sub(%r{\A[^/]*/}, ''),
+                              path: path}
+      end
+      it { expect(response).to have_http_status(:ok) }
+      it { |example| expect([example, response]).to comply_with_rest_api }
+      it 'finds the repository' do
+        expect(response_data['repository']).not_to be(nil)
+      end
```
eugenk: Please add a test if the requested log contains all commits. You can get a list of all commits (Gitlab::Git::Commit objects) that change the file at `path` with
```ruby
repository.git.log(path: path)
```
If you are interested in the commit hashes instead, use
```ruby
repository.git.log(path: path).map(&:id)
```

Compare this to the commit ids that you receive from the REST API.

spec/support/api/schemas/rest/history_controller.json
```diff
@@ -0,0 +1,84 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "title": "Rest Histories",
```
eugenk: "REST History"

PR #240, opened 2017-10-13
239 add permissions field to query current users permissions

### LINKED ISSUE #239: Add permissions field to query current_user's permissions
This should be added to `OrganizationType` and `RepositoryType`.

phyrog:
Closes #239.

### DIFF
--- a/spec/graphql/types/repository/permissions_type_spec.rb
+++ b/spec/graphql/types/repository/permissions_type_spec.rb
@@ -1,0 +1,52 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe Types::Repository::PermissionsType do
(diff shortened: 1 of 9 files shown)


spec/graphql/types/repository/permissions_type_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+  let!(:user) { create :user }
+
+  subject { repository }
+
+  let(:permissions_type) do
+    OntohubBackendSchema.types['RepositoryPermissions']
+  end
+
+  before do
+    subject.add_member(user, 'admin')
```
eugenk: Please test for something other than 'admin' at this point. The current test would also allow an implementation that constantly returns 'admin'.

PR #257, opened 2017-10-21
Post process analysis jobs

eugenk:
This is not yet ready for review. There is one issue I cannot solve and I need help:

The [specs of the ImportingDocumentsReanalyzer](https://github.com/ontohub/ontohub-backend/blob/796a9f6b212b99a8e5c05ca8af763e1b37f52570/spec/lib/importing_documents_reanalyzer_spec.rb#L148-L205) are failing randomly. Even if only one of them is run. Sometimes, it's green and sometimes, it's red.

Apart from this problem:

This finalises the evaluation pipeline. Most of the lines of code are fixtures from Hets-lib. These casl files don't need to be reviewed. Most important, this adds post-processing to analysis jobs:

* A file is saved in a repository
* The backend pushes a job to the HetsAgent
* The HetsAgent runs the job and pushes a job back to the backend (https://github.com/ontohub/hets-agent/pull/38)
* The Backend finds the Documents that import the just analysed Document and invokes the analysis of these (this PR)

This PR depends on https://github.com/ontohub/gitlab_git/pull/20 and https://github.com/ontohub/hets-agent/pull/38

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -11,7 +11,7 @@ GIT
 
 GIT
   remote: git://github.com/ontohub/ontohub-models.git
-  revision: 47b9b0f5fafdf676fb9f54b3d62b08a2574f915e
+  revision: c5dab310b1000f899bce9100f3cf73afcfdfe717
   branch: master
   specs:
     ontohub-models (0.1.0)
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -11,7 +11,7 @@ GIT
 
 GIT
   remote: git://github.com/ontohub/ontohub-models.git
-  revision: c5dab310b1000f899bce9100f3cf73afcfdfe717
+  revision: 47b9b0f5fafdf676fb9f54b3d62b08a2574f915e
   branch: master
   specs:
     ontohub-models (0.1.0)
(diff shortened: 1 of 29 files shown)


phyrog:
I think the problem is the document factory, after poking around a bit, I found this:

```ruby
d.kind
=> "NativeDocument"
Document.first(file_version_id: 15501, kind: d.kind)
=> nil
Document.first(file_version_id: 15501)
=> #<Library @values={:id=>1529, :file_version_id=>15501, :kind=>"Library", :loc_id=>"ver ad canonicus spargo", :created_at=>2017-10-24 13:00:47 +0200, :updated_at=>2017-10-24 13:00:47 +0200, :display_name=>"desipio apparatus demonstro casso", :name=>"desipio-apparatus-demonstro-casso", :location=>"http://schuster.name/angela", :version=>"2.46"}>
```

(See [ImportingDocumentsReanalyzer](https://github.com/ontohub/ontohub-backend/blob/796a9f6b212b99a8e5c05ca8af763e1b37f52570/lib/importing_documents_reanalyzer.rb#L63-L64))

So the previous document version has a different kind than the current one, which causes `process` to skip the analysis of the importers of the previous version and thus not updating the FileVersionParents

eugenk:
Thanks! The implementation should be fixed now. I added a commit and rebased on master. I also filed https://github.com/ontohub/ontohub-models/pull/118 to prevent such things from happening.

PR #267, opened 2017-10-28
Create GraphQL REST controller for Git Diff.

### LINKED ISSUE #188: Create GraphQL-REST Controller for Git Diff
# Diff
## Action `single_commit`
* route
  The controller action is supposed to show the git diff of the revision or revision range. If no revision is supplied by the user, use the default branch.
  * `/v3/:organizational_unit_id/:repository_id/diff/:revision` 
      Diff of a single commit. This needs to call `repository.commit(revision: params[:revision]).diff` in GraphQL.

## Action `commit_range`
* route
  * `/v3/:organizational_unit_id/:repository_id/diff/:revision_from..:revision_to/*path`
      Diff of a commit range in an optional path (the route is accepted if the path is empty because of the wildcard segment). This needs to call `repository.diff(from: params[:revision_from], to: params[:revision_to], paths: [params[:path]])` in GraphQL.
     This is tricky because we need to forbid two consecutive dots `..` in the revisions *and* it must not start with a dot`.`. Use a `constraint` on the two revisions with this regex: `REVISION = ([^\.]+)(\.[^\.]+)*([^\.]+)` analogous to `UNTIL_DOUBLE_SLASHES` in the routes.rb.

## Common in both actions
* fields
  * `deletedFile`
  * `diff`
  * `lineCount`
  * `newFile`
  * `newMode`
  * `newPath`
  * `oldMode`
  * `oldPath`
  * `renamedFile`

eugenk:
This shall fix #188.

### DIFF
--- a/spec/support/api/schemas/rest/diffs_controller.json
+++ b/spec/support/api/schemas/rest/diffs_controller.json
@@ -13,27 +13,34 @@
             "required": ["repository"],
             "properties": {
               "repository": {
-                "type": "object",
-                "required": ["commit"],
-                "properties": {
-                  "commit": {
-                    "oneOf": [
-                      {
-                        "type": "object",
-                        "required": ["diff"],
-                        "properties": {
-                          "diff": {
-                            "type": "array",
-                            "items": {"$ref": "#/definitions/diff"}
+                "oneOf": [
+                  {
+                    "type": "object",
+                    "required": ["commit"],
+                    "properties": {
+                      "commit": {
+                        "oneOf": [
+                          {
+                            "type": "object",
+                            "required": ["diff"],
+                            "properties": {
+                              "diff": {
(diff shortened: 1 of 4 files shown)


spec/support/api/schemas/rest/diffs_controller.json
```diff
@@ -0,0 +1,195 @@
...
+    "actions": {
+      "single_commit": {
+        "type": "object",
+        "required": ["data"],
+        "properties": {
+          "data": {
+            "type": "object",
+            "required": ["repository"],
+            "properties": {
+              "repository": {
```
DerProfessor: please add a oneOf for repository because it can be null

eugenk:
I rebased on master and added a commit 484573d to incorporate your requested changes.

PR #278, opened 2017-11-03
250 graphql add types and mutations for urlmapping

### LINKED ISSUE #250: GraphQL: Add types and mutations for UrlMapping
A `Repository` can have many `UrlMapping`s. 

A `UrlMapping` contains
* `id`: Integer, not null
* `number`: Integer, not null
* `source`: String, not null
* `target`: String, not null

#### Type
Add a field `urlMappings` to `Repository` that returns a list of `UrlMapping`s.
Files to use:
* Edit the file [app/graphql/types/repository_type.rb](https://github.com/ontohub/ontohub-backend/blob/6f675c52ecc3fdb3221362886b645801830960cb/app/graphql/types/repository_type.rb) to add the field
* Create a file app/graphql/types/repository/url_mapping_type.rb to add the type of the field.
* Add tests for the field to [spec/graphql/types/repository_type_spec.rb](https://github.com/ontohub/ontohub-backend/blob/6f675c52ecc3fdb3221362886b645801830960cb/spec/graphql/types/repository_type_spec.rb). You don't need to test the new type because it does not contain logic to resolve its fields.

#### Mutation: Add
Add a mutation `addUrlMapping` that creates a `UrlMapping` in a repository. 
Files to use:
* Create a file app/graphql/types/repository/url_mapping/input_type.rb to add the type of mutation's input. The fields `id` and `number` don't need to be set here. They are set automatically by the database. The arguments needed are: `repositoryId`, `source`, `target`.
* Create a file app/graphql/mutations/repository/add_url_mapping_mutation.rb that actually creates the `UrlMapping`. You can create a UrlMapping with 
  ```ruby
    UrlMapping.create(repository: Repository.first(slug: arguments['repositoryId']),
                      source: arguments['source'],
                      target: arguments['target'])
  ```
* Edit the file [app/graphql/types/mutation_type.rb](https://github.com/ontohub/ontohub-backend/blob/6f675c52ecc3fdb3221362886b645801830960cb/app/graphql/types/mutation_type.rb) to register the new mutation.
* Create a file spec/graphql/mutations/repository/add_url_mapping_mutation_spec.rb with tests for this mutation.

#### Mutation: Remove
Add a mutation `removeUrlMapping` that removes a `UrlMapping` by its `id`. 
Files to use:
* Create a file app/graphql/mutations/repository/remove_url_mapping_mutation.rb that finds the `UrlMapping` inside the repository and deletes it. This should lead to an error if the `id` does not exist *inside the repository*.
* Edit the file [app/graphql/types/mutation_type.rb](https://github.com/ontohub/ontohub-backend/blob/6f675c52ecc3fdb3221362886b645801830960cb/app/graphql/types/mutation_type.rb) to register the new mutation.
* Create a file spec/graphql/mutations/repository/remove_url_mapping_mutation_spec.rb with tests for this mutation.

## Orientation
You can get some orientation on how to do this by looking at the implementation of querying/adding/removing public keys of a user:
[app/graphql/types/public_key_type.rb](https://github.com/ontohub/ontohub-backend/blob/6f675c52ecc3fdb3221362886b645801830960cb/app/graphql/types/public_key_type.rb),
[app/graphql/mutations/account/add_public_key_mutation.rb](https://github.com/ontohub/ontohub-backend/blob/6f675c52ecc3fdb3221362886b645801830960cb/app/graphql/mutations/account/add_public_key_mutation.rb) and
[app/graphql/mutations/account/remove_public_key_mutation.rb](https://github.com/ontohub/ontohub-backend/blob/6f675c52ecc3fdb3221362886b645801830960cb/app/graphql/mutations/account/remove_public_key_mutation.rb).

DerProfessor:
Closes #250

### DIFF
--- a/app/graphql/mutations/repository/remove_url_mapping_mutation.rb
+++ b/app/graphql/mutations/repository/remove_url_mapping_mutation.rb
@@ -15,7 +15,7 @@ module Mutations
       end
 
       resource!(lambda do |_root, arguments, _context|
-        ::RepositoryCompound.first(slug: arguments[:repositoryId])
+        RepositoryCompound.first(slug: arguments[:repositoryId])
       end)
 
       authorize! :update
--- a/app/graphql/mutations/repository/remove_url_mapping_mutation.rb
+++ b/app/graphql/mutations/repository/remove_url_mapping_mutation.rb
@@ -1,0 +1,43 @@
+# frozen_string_literal: true
+
+module Mutations
+  module Repository
+    RemoveUrlMappingMutation = GraphQL::Field.define do
+      type types.Boolean
+      description 'Removes an URL Mapping'
+
+      argument :repositoryId, !types.ID do
+        description 'The ID of the repository in which the URL mapping exits'
+      end
+
+      argument :urlMappingId, !types.ID do
+        description 'The ID of the URL mapping'
+      end
+
+      resource!(lambda do |_root, arguments, _context|
+        ::Repository.first(slug: arguments[:repositoryId])
+      end)
+
+      authorize! :update
+
+      resolve RemoveUrlMappingResolver.new
+    end
+
+    # GraphQL mutation to add new URL mappings
+    class RemoveUrlMappingResolver
+      def call(repository, arguments, context)
+        url_mapping = repository.url_mappings_dataset.
+          first(id: arguments[:urlMappingId])
+        if url_mapping.nil?
+          context.add_error(GraphQL::ExecutionError.new('resource not found'))
+
+          false
+        else
+          url_mapping.destroy
+
+          true
+        end
+      end
+    end
+  end
+end
--- a/app/graphql/mutations/repository/remove_url_mapping_mutation.rb
+++ b/app/graphql/mutations/repository/remove_url_mapping_mutation.rb
@@ -15,7 +15,7 @@ module Mutations
       end
 
       resource!(lambda do |_root, arguments, _context|
-        ::Repository.first(slug: arguments[:repositoryId])
+        ::RepositoryCompound.first(slug: arguments[:repositoryId])
       end)
 
       authorize! :update
--- a/spec/graphql/mutations/repository/remove_url_mapping_mutation_spec.rb
+++ b/spec/graphql/mutations/repository/remove_url_mapping_mutation_spec.rb
@@ -62,7 +62,7 @@ RSpec.describe Mutations::Repository::RemoveUrlMappingMutation do
           expect(subject['data']['removeUrlMapping']).
             to eq(true)
         end
-    
+
         it 'removes the mapping from the database' do
           subject
           expect(UrlMapping.first(id: url_mapping.id)).
--- a/spec/graphql/mutations/repository/remove_url_mapping_mutation_spec.rb
+++ b/spec/graphql/mutations/repository/remove_url_mapping_mutation_spec.rb
@@ -1,0 +1,140 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe Mutations::Repository::RemoveUrlMappingMutation do
+  let!(:user) { create :user }
+  let(:url_mapping) { create(:url_mapping) }
+  let(:repository) { url_mapping.repository }
+  before do
+    repository.add_member(user, 'admin')
+  end
+
+  let(:context) { {current_user: current_user} }
+
+  let(:url_mapping_id) { url_mapping.to_param }
+  let(:repository_id) { repository.to_param }
+  let(:variables) do
+    {'urlMappingId' => url_mapping_id,
+     'repositoryId' => repository_id}
+  end
+
+  let(:result) do
+    OntohubBackendSchema.execute(
+      query_string,
+      context: context,
+      variables: variables
+    )
+  end
+
+  let(:query_string) do
+    <<-QUERY
+    mutation ($repositoryId: ID!, $urlMappingId: ID!) {
+      removeUrlMapping(repositoryId: $repositoryId, urlMappingId: $urlMappingId)
+    }
+    QUERY
+  end
+
+  subject { result }
+
+  shared_examples 'handling the deletion' do
+    it 'deletes Url Mapping' do
+      expect(subject['data']['removeUrlMapping']).
+        to eq(expectation_response_matcher)
+    end
+
+    it 'deletes Url Mapping, which already has been deleted' do
+      subject
+      expect(UrlMapping.first(id: url_mapping.id)).
+        to be(expectation_database_object)
+    end
+  end
+
+  shared_examples 'failing to delete' do
+    it 'deletion of Url Mapping failed' do
+      expect(subject['data']['removeUrlMapping']).
+        to eq(expectation_response_matcher)
+    end
+
+    it 'deletes Url Mapping, which already has been deleted' do
+      subject
+      expect(UrlMapping.first(id: url_mapping.id)).
+        to eq(url_mapping)
+    end
+
+    it 'error mesage when deletion faild' do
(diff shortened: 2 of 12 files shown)


app/graphql/mutations/repository/remove_url_mapping_mutation.rb
```diff
@@ -0,0 +1,43 @@
...
+      argument :repositoryId, !types.ID do
+        description 'The ID of the repository in which the URL mapping exits'
+      end
+
+      argument :urlMappingId, !types.ID do
+        description 'The ID of the URL mapping'
+      end
+
+      resource!(lambda do |_root, arguments, _context|
+        ::Repository.first(slug: arguments[:repositoryId])
```
phyrog: Please use `RepositoryCompound` instead.

app/graphql/mutations/repository/remove_url_mapping_mutation.rb
```diff
@@ -0,0 +1,43 @@
...
+
+      resolve RemoveUrlMappingResolver.new
+    end
+
+    # GraphQL mutation to add new URL mappings
+    class RemoveUrlMappingResolver
+      def call(repository, arguments, context)
+        url_mapping = repository.url_mappings_dataset.
+          first(id: arguments[:urlMappingId])
+        if url_mapping.nil?
```
phyrog: I think this is not needed. Since the goal is to remove the mapping, if the mapping wasn't there in the first place, we still achieved what we wanted.
You can just do
```ruby
url_mapping.destroy if url_mapping
true
```

spec/graphql/mutations/repository/remove_url_mapping_mutation_spec.rb
```diff
@@ -0,0 +1,140 @@
...
+    mutation ($repositoryId: ID!, $urlMappingId: ID!) {
+      removeUrlMapping(repositoryId: $repositoryId, urlMappingId: $urlMappingId)
+    }
+    QUERY
+  end
+
+  subject { result }
+
+  shared_examples 'handling the deletion' do
+    it 'deletes Url Mapping' do
```
phyrog: If we always return true, this test is not needed.

spec/graphql/mutations/repository/remove_url_mapping_mutation_spec.rb
```diff
@@ -0,0 +1,140 @@
...
+      expect(subject['data']['removeUrlMapping']).
+        to eq(expectation_response_matcher)
+    end
+
+    it 'deletes Url Mapping, which already has been deleted' do
```
phyrog: This sounds like it's deleting something again, but you're checking if the mapping has been deleted properly. So maybe `'removes the mapping from the database'`.

spec/graphql/mutations/repository/remove_url_mapping_mutation_spec.rb
```diff
@@ -0,0 +1,140 @@
...
+      subject
+      expect(UrlMapping.first(id: url_mapping.id)).
+        to be(expectation_database_object)
```
phyrog: If the deletion was successful, we always expect it to return `nil` afterwards, so you can just hardcode it here instead of passing it via `expectation_database_object`.

spec/graphql/mutations/repository/remove_url_mapping_mutation_spec.rb
```diff
@@ -0,0 +1,140 @@
...
+    end
+  end
+
+  shared_examples 'failing to delete' do
+    it 'deletion of Url Mapping failed' do
```
phyrog: See above, we should always return `true`.

spec/graphql/mutations/repository/remove_url_mapping_mutation_spec.rb
```diff
@@ -0,0 +1,140 @@
...
+      expect(subject['data']['removeUrlMapping']).
+        to eq(expectation_response_matcher)
+    end
+
+    it 'deletes Url Mapping, which already has been deleted' do
+      subject
+      expect(UrlMapping.first(id: url_mapping.id)).
```
phyrog: You could just check for `.not_to be(nil)`, so you don't have to pass the actual object.

spec/graphql/mutations/repository/remove_url_mapping_mutation_spec.rb
```diff
@@ -0,0 +1,140 @@
...
+    end
+  end
+
+  shared_examples 'failing to delete' do
+    it 'deletion of Url Mapping failed' do
+      expect(subject['data']['removeUrlMapping']).
+        to eq(expectation_response_matcher)
+    end
+
+    it 'deletes Url Mapping, which already has been deleted' do
```
phyrog: `'does not remove the database entry'`

spec/graphql/mutations/repository/remove_url_mapping_mutation_spec.rb
```diff
@@ -0,0 +1,140 @@
...
+      subject
+      expect(UrlMapping.first(id: url_mapping.id)).
+        to eq(url_mapping)
+    end
+
+    it 'error mesage when deletion faild' do
```
phyrog: `'returns an error message when deletion failed'`

spec/graphql/mutations/repository/remove_url_mapping_mutation_spec.rb
```diff
@@ -0,0 +1,140 @@
...
+        it_behaves_like 'handling the deletion' do
+          let(:expectation_response_matcher) { true }
+          let(:expectation_database_object) { nil }
+        end
+      end
+      context 'When the object does not exist' do
+        it_behaves_like 'failing to delete' do
+          let(:url_mapping_id) { 0 }
+          let(:expectation_response_matcher) { false }
+          let(:error_message) { 'resource not found' }
```
phyrog: If the user is authorized, we should not return an error if the mapping did not exist.

spec/graphql/mutations/repository/remove_url_mapping_mutation_spec.rb
```diff
@@ -0,0 +1,140 @@
...
+        end
+      end
+    end
+    context 'Repository ID is not vaild' do
+      let(:repository_id) { "bad-#{repository.to_param}" }
+      context 'When the URL mapping exists' do
+        it_behaves_like 'failing to delete' do
+          let(:expectation_response_matcher) { nil }
+          let(:error_message) { 'resource not found' }
```
phyrog: If the user is authorized, we should not return an error if the mapping did not exist.

spec/graphql/mutations/repository/remove_url_mapping_mutation_spec.rb
```diff
@@ -0,0 +1,140 @@
...
+        end
+      end
+      context 'When the object does not exist' do
+        it_behaves_like 'failing to delete' do
+          let(:url_mapping_id) { 0 }
+          let(:expectation_response_matcher) { nil }
+          let(:error_message) { 'resource not found' }
```
phyrog: If the user is authorized, we should not return an error if the mapping did not exist.

app/graphql/mutations/repository/remove_url_mapping_mutation.rb
```diff
@@ -15,7 +15,7 @@ module Repository
       end
 
       resource!(lambda do |_root, arguments, _context|
-        ::Repository.first(slug: arguments[:repositoryId])
+        ::RepositoryCompound.first(slug: arguments[:repositoryId])
```
phyrog: I think you can remove the `::` now (It was only needed because we are also in a Module named `Repository`, so it didn't know which `Repository` we meant.

PR #279, opened 2017-11-05
277 dont leak existence of a private repository

### LINKED ISSUE #277: Don't leak existence of a private repository
Consider the query
```graphql
query( $id: !ID) {
  repository(id: $id) {
    name
  }
}
```

Setting the `$id` to `"the-league-of-extraordinary-users/top-secret"` while signed in as ada, the result is:
```
{
  "data": {
    "repository": {
      "name": "Top Secret"
    }
  }
}
```
which is just fine. However, when not signed in, the result is:
```
{
  "data": {
    "repository": null
  }
}
```
It looks like the message that we want to have when a repository doesn't exist, but it isn't. When setting the `$id` to `"i-dont-exist"`, the result is
```
{
  "data": {
    "repository": null
  },
  "errors": [
    {
      "message": "resource not found",
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ],
      "path": [
        "repository"
      ]
    }
  ]
}
```

The last two results must be the same if we don't want to leak the existence of a private repository.

phyrog:
Fixes #277. At least in part.

There are at least two other places where we leak the existence of private repositories:

- `DeleteRepositoryMutation`
- `SaveRepositoryMutation`

These both return an error if the repository does not exist, but a different error if the user is unauthorized to perform the mutations. I'm unclear about how to solve this: how do we check if a user is authorized to see if a non-existent repository is not existent? Returning `unauthorized` even if the repo does not exist? Could be confusing if there's just a typo in the name.

### DIFF
--- a/app/policies/repository_policy.rb
+++ b/app/policies/repository_policy.rb
@@ -22,6 +22,7 @@ class RepositoryPolicy < ApplicationPolicy
   end
 
   def show?
+    return false unless resource
     resource.public_access ||
       !!current_user&.accessible_repositories_dataset&.
         where(slug: resource.to_param)&.any?
--- a/app/graphql/instrumenters/not_found_unless_instrumenter.rb
+++ b/app/graphql/instrumenters/not_found_unless_instrumenter.rb
@@ -1,0 +1,33 @@
+# frozen_string_literal: true
+
+module Instrumenters
+  # Rescues Sequel::ValidationFailed errors and raises
+  # GraphQL::ExecutionError instead
+  class NotFoundUnlessInstrumenter
+    # rubocop:disable Metrics/MethodLength
+    # rubocop:disable Metrics/AbcSize
+    def instrument(_type, field)
+      # rubocop:enable Metrics/MethodLength
+      # rubocop:enable Metrics/AbcSize
+      query = field.metadata[:not_found_unless]
+      return field unless query
+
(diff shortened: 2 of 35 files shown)


eugenk:
We should return the equivalent of 404 in both cases: if we don't have access and if the repository does not exist.

eugenk:
There are these cases and the desired "return codes" (or at least their equivalents):
* user has admin access to the repository -> 200
* user has read or write access to the repository -> 403
* user has no read access to the repository -> 404
* repository does not exist -> 404

eugenk:
I added a test in https://github.com/ontohub/ontohub-backend/pull/276/commits/8ecd14e36ce7a1e9c121636cd71111225fafa3eb. It is red [in that branch](https://github.com/ontohub/ontohub-backend/pull/276), but it tests this expected behaviour.

phyrog:
I guess we'll have to change the `resource`s to only return the repo if a user has at least read access.

app/policies/repository_policy.rb
```diff
@@ -22,6 +22,7 @@ def resolve
   end
 
   def show?
+    return false unless resource
```
eugenk: Please add a (policy) spec for this case.

phyrog: Done in b23173f

PR #291, opened 2017-11-12
181 Add API keys

### LINKED ISSUE #181: Add API keys
In order for Hets to fetch data from private repositories, we should add API keys.

Since Hets only needs to *read* repository data, it suffices to change the `RepositoryPolicy#show?` method to check for the API key.

---

An API key should be handled the same way as a Devise reset password token, such that the key itself is not saved in the database, but rather a digest value.

Devise token generation: [[1]](https://github.com/plataformatec/devise/blob/ee01bac8b0b828b3da0d79c46115ba65c433d6c8/lib/devise/models/recoverable.rb#L87-L94) [[2]](https://github.com/plataformatec/devise/blob/88724e10adaf9ffd1d8dbfbaadda2b9d40de756a/lib/devise/token_generator.rb#L17-L20)
Devise token validation: [[3]](https://github.com/plataformatec/devise/blob/ee01bac8b0b828b3da0d79c46115ba65c433d6c8/lib/devise/models/recoverable.rb#L144-L147) [[4]](https://github.com/plataformatec/devise/blob/88724e10adaf9ffd1d8dbfbaadda2b9d40de756a/lib/devise/token_generator.rb#L10-L12)

eugenk:
Closes #181. Depends on https://github.com/ontohub/ontohub-models/pull/125.

The policy specs all remain the same except for indentation and the addition of a new context at the bottom.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -11,7 +11,7 @@ GIT
 
 GIT
   remote: git://github.com/ontohub/ontohub-models.git
-  revision: 0f279f0a7734da0a42c205fe0728dcaade6caef8
+  revision: c8482f769f7a1e0de0111c5b6e1ce2a36ffb8ca4
   branch: master
   specs:
     ontohub-models (0.1.0)
@@ -431,4 +431,4 @@ DEPENDENCIES
   sprockets-rails
 
 BUNDLED WITH
-   1.15.4
+   1.16.0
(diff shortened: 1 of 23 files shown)


eugenk:
Done. Rebased on master and moved the "Update ontohub-models." commit to the top.

Issue #96, opened 2017-04-12
Git: Cloning


eugenk:
When creating a repository, allow to clone a remote git/svn repository. Either as a fork that is writable or as a mirror that is synchronised periodically and write-protected. When creating such a mirror/fork, allow to specify a list of UrlMappings.

This depends on https://github.com/ontohub/ontohub-models/issues/114 and https://github.com/ontohub/ontohub-backend/issues/250 

Use `GitHelper.exclusively(repository) { repository.git.pull }` to wrap the synchronisation process in a mutex. See #85 and #304.

# Implementation Hints
## Cloning
* Create a GraphQL mutation 
    ```graphql
    mutation ($newRepository: NewRepository!, $remoteAddress: String!, $remoteType: RepositoryRemoteTypeEnum!, $urlMappings: [UrlMapping!]!) : Repository {
      cloneRepository(data: $newRepository, remoteAddress: $remoteAddress, type: $type, urlMappings: $urlMappings)
    }
    ```
    that creates a `Repository` (not a `RepositoryCompound`) similar to the `createRepositoryMutation` but also sets the additional fields appropriately. After creating, it schedules an asynchronous background job with `RepositoryCloningJob.perform_later(repository.id)`.
* **Edit**: Before creating the Repository, call `Bringit::Wrapper.valid_remote?(remote_address)` to check whether or not the remote is a supported repository. If it returns `false`, add an error to the GraphQL context.
* Create a `RepositoryCloningJob` with a public method `perform(repository_id)` that fetches the Repository from the database and [clones the git repository with gitlab_git](https://github.com/ontohub/gitlab_git#clone). Right after cloning, the `synchronized_at` timestamp should be set to the current time.
* Create a `RepositoryCloningWorker` that only sets the options (queue, threads, prefetch, timeout) appropriately. See, for instance, the `PostProcessHetsWorker` as an example.
* Add a simple, small bare git repository and an svn repository to `db/seeds/fixtures/repositories`. Clone these two repositories at the end of `db/seeds/030_repository_seeds.rb` the same way you clone them in the `createRepositoryMutation`.
## Pulling
* Create a `RepositoryPullingWorker` that only sets the appropriate options.
[...]

PR #304, opened 2017-11-18
85 git mutex git actions

### LINKED ISSUE #85: Git: Mutex git actions
We use two different systems to write to git repositories: [git-shell](https://github.com/ontohub/ontohub-git-shell) and [gitlab-git](https://github.com/ontohub/gitlab_git). The write actions need to be done in a critical path, exclusively for one process. Use an appropriate mutex mechanism to ensure that only one process has write access. Lock-files may be a good option.

eugenk:
This shall fix #85 at least for the current implementation. It needs to be used for git pull and the git shell as well. I already added a note about this to the cloning issue #96.

### DIFF
--- a/app/models/authorized_keys_file.rb
+++ b/app/models/authorized_keys_file.rb
@@ -10,7 +10,7 @@ class AuthorizedKeysFile
         "#{key.key.strip} #{key.name}"
       end
 
-      FileLockHelper.exclusively(LOCK_FILE, timeout: 5) do
+      FileLockHelper.exclusively(LOCK_FILE, timeout: 5.seconds) do
         path = Rails.root.join('tmp/data/authorized_keys')
         path.dirname.mkpath
         File.write(path, keys.join("\n"))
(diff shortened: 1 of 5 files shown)


app/models/authorized_keys_file.rb
```diff
@@ -2,23 +2,19 @@
...
+  LOCK_FILE = 'authorized_keys.lock'
+
   class << self
     def write
       keys = PublicKey.all.map do |key|
         "#{key.key.strip} #{key.name}"
       end
 
-      Filelock(lock_file, timeout: 5) do
+      FileLockHelper.exclusively(LOCK_FILE, timeout: 5) do
```
phyrog: Can you change the `5` to `5.seconds`, just to be more explicit about it?

eugenk: Done in 6ec95ce7

eugenk:
I added a commit to use an in-repository lock file. With this change, the Git-Shell does not need to know the location of the backend's lock files.

PR #319, opened 2017-11-28
Improve boot time.

eugenk:
[Bootsnap](https://github.com/Shopify/bootsnap) is a gem that reduces the time to boot an application by changing how `require` operates. This gem is [included in the default Gemfile in Rails 5.2](http://weblog.rubyonrails.org/2017/11/27/Rails-5-2-Active-Storage-Redis-Cache-Store-HTTP2-Early-Hints-Credentials/). In a first test, we save about a second compared to the current master branch (according to the RSpec report).


eugenk:
Rebased on master and solved Gemfile.lock conflict.

PR #335, opened 2017-12-20
96 git cloning

### LINKED ISSUE #96: Git: Cloning

DerProfessor:
Fixes #96


spec/graphql/mutations/repository/clone_repository_mutation_spec.rb
```diff
@@ -0,0 +1,97 @@
...
+
+  context 'remote is invalid' do
+    before do
+      # Stub the Bringit::Wrapper
+      allow(Bringit::Wrapper).
+        to receive(:valid_remote?).
+        and_return(false)
+    end
+
+    it 'enqueues a clone-repository job' do
```
eugenk: does not enqueue a clone-repository job

spec/graphql/mutations/repository/clone_repository_mutation_spec.rb
```diff
@@ -0,0 +1,97 @@
...
+      expect(RepositoryCloningJob).
+        not_to have_been_enqueued
+    end
+
+    it 'error has been added to the graphql context' do
```
eugenk: adds a GraphQL error

spec/graphql/mutations/repository/clone_repository_mutation_spec.rb
```diff
@@ -0,0 +1,97 @@
...
+        contentType
+        visibility
+      }
+    }
+    QUERY
+  end
+
+  subject { result }
+
+  context 'remote is valid' do
```
eugenk: with a valid remote

spec/graphql/mutations/repository/clone_repository_mutation_spec.rb
```diff
@@ -0,0 +1,97 @@
...
+        and_return(true)
+    end
+    it 'enqueues a clone-repository job' do
+      expect(RepositoryCloningJob).
+        to have_been_enqueued.
+        with(subject['data']['cloneRepository']['id'])
+    end
+  end
+
+  context 'remote is invalid' do
```
eugenk: with an invalid remote

app/graphql/mutations/repository/clone_repository_mutation.rb
```diff
@@ -0,0 +1,63 @@
...
+
+      argument :remoteAddress, !types.String do
+        description 'The addres of a remote repository'
+      end
+
+      argument :remoteType, !Types::Repository::RepositoryRemoteTypeEnum do
+        description 'The type of a remote repository'
+      end
+
+      argument :newUrlMappings, !types[!Types::Repository::NewUrlMappingType] do
```
eugenk: Let's name this argument `urlMappings`.

app/graphql/types/repository/repository_remote_type_enum.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+Types::Repository::RepositoryRemoteTypeEnum = GraphQL::EnumType.define do
+  name 'RepositoryRemoteTypeEnum'
```
eugenk: Let's change the name to `"RepositoryRemoteType"`

app/graphql/mutations/repository/clone_repository_mutation.rb
```diff
@@ -0,0 +1,63 @@
...
+        params['public_access'] = params.delete('visibility') == 'public'
+
+        if Bringit::Wrapper.valid_remote?(arguments[:remote_address])
+          repository = ::Repository.create(params)
+          RepositoryCloningJob.perform_later(repository.to_param)
+          repository
+        else
+          add_invalid_remote_error(context, arguments[:remote_address])
+        end
+      end
```
eugenk: `:remote_address` should be `'remoteAddress'`
The same applies to `remoteType` which is not even used. You can do this with:

```ruby
params.merge!(arguments.to_h.slice('remoteAddress', 'remoteType').transform_keys { |k| k.to_s.underscore })
# or
params['remote_address'] = arguments['remoteAddress']
params['remote_type'] = arguments['remoteType']
```
after the line `params['public_access'] = ...`.

Also, the `UrlMapping`s are not created:

```ruby
arguments['urlMappings'].each do |url_mapping|
  UrlMapping.create(repository_id: repository.id, source: url_mapping['source'], target: url_mapping['target'])
end
```

spec/jobs/repository_pulling_periodically_job_spec.rb
```diff
@@ -0,0 +1,41 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe RepositoryPullingPeriodicallyJob, type: :job do
+  let!(:repository_mirror1) { create :repository, :mirror }
+  let!(:repository_mirror2) { create :repository, :mirror }
+  let!(:repository_fork) { create :repository, :fork }
+
+  it 'sets the time for the job to be scheduled' do
```
eugenk: `it 're-enqueues itself' do`

spec/graphql/mutations/repository/clone_repository_mutation_spec.rb
```diff
@@ -59,39 +59,75 @@
...
       expect(subject.to_h).
         to match('data' => {'cloneRepository' => nil},
                  'errors' =>
                    [include('message' =>
                      /remote_address: ".*" is not a git or svn repository/)])
     end
+
+    it 'repository was not created' do
+      expect(UrlMapping).
+        not_to have_received(:create)
```
eugenk: This is a way to check if the repository has been created. A more intuitive way would be:
```ruby
expect(Repository.first(slug: repository_blueprint.to_param)).to be(nil)
```
... because you check the existence of the *repository* instead of the url mapping.

app/graphql/mutations/repository/clone_repository_mutation.rb
```diff
@@ -0,0 +1,75 @@
...
+      authorize!(lambda do |owner, _arguments, context|
+        RepositoryPolicy.new(context[:current_user], nil).create?(owner)
+      end)
+
+      resolve CloneRepositoryResolver.new
+    end
+
+    # GraphQL mutation to clone a new repository
+    class CloneRepositoryResolver
+      def call(_owner, arguments, context)
```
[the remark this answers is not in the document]
DerProfessor: I transfered a code block into a new function called `create_repository`. Because the line of code has not changed, hound does not recognize the changes.

Issue #343, opened 2018-01-01
Replace SettingsValidator with dry-validation


eugenk:
Replace the [SettingsValidator's custom validations](https://github.com/ontohub/ontohub-backend/blob/b1c461eb6c7eaad91a4904b8ad671d93541f93da/config/initializers/settings_validator.rb) with [dry-validation](http://dry-rb.org/gems/dry-validation/). See especially [the validation section from config's readme](https://github.com/railsconfig/config/tree/1.6.1#validation) for details.

PR #349, opened 2018-01-07
Add RabbitMQ prefix and exchange options

eugenk:
This shall add prefix and exchange options for RabbitMQ. Queue names will be prefixed and the exchange will be chosen by the app. This is needed for the system-tests to not interfere with the backend seeds because the queues are emptied in the seeds (and they run in the system-tests).

This pull request is blocked because of #343: It still needs to validate the settings.

### DIFF
--- a/app/jobs/post_process_hets_job.rb
+++ b/app/jobs/post_process_hets_job.rb
@@ -2,7 +2,7 @@
 
 # Post-processes a finished job that was sent to Hets
 class PostProcessHetsJob < ApplicationJob
-  queue_as :post_process_hets
+  queue_as "#{Settings.rabbitmq.prefix}_post_process_hets"
 
   def perform(result, original_job_message)
     return unless result == 'success'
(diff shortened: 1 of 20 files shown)


eugenk:
ping

phyrog:
Once the new settings key is validated :+1:

eugenk:
Well, we have a circular dependency:

settings_schema.rb:
```ruby
Dir.glob('app/workers/*.rb').sort.each do |file|
  require_relative Rails.root.join(file).to_s
end
```

application_worker.rb:
```ruby
class ApplicationWorker
  ...  
  from_queue "#{Settings.rabbitmq.prefix}_default"
```

phyrog:
Yikes. We could possibly hack something into `settings_schema.rb` to not load the files, but look at the file name and define an empty class for each of the files which are reopened when loading the worker files.

eugenk:
I think that's too complicated. A simple list of allowed values (maintained by us) should suffice. There won't be that many workers.

eugenk:
Done. It's ready for review

PR #354, opened 2018-01-10
Add missing field: timeLimit

eugenk:
This adds a field to the API that is already present in the models, but was not exposed to the end user.

### DIFF
--- a/app/graphql/types/language_mapping_type.rb
+++ b/app/graphql/types/language_mapping_type.rb
@@ -4,7 +4,7 @@ Types::LanguageMappingType = GraphQL::ObjectType.define do
   name 'LanguageMapping'
   description 'A mapping between two languages'
 
-  field :id, !types.ID do
+  field :id, !types.Int do
     description 'The ID of this LanguageMapping'
   end
 
--- a/app/graphql/types/query_type.rb
+++ b/app/graphql/types/query_type.rb
@@ -73,7 +73,7 @@ Types::QueryType = GraphQL::ObjectType.define do
   field :languageMapping, Types::LanguageMappingType do
     description 'A LanguageMapping for the given ID'
 
-    argument :id, !types.ID do
+    argument :id, !types.Int do
       description 'The ID of the LangaugeMapping'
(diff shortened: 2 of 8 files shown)


app/graphql/types/language_mapping_type.rb
```diff
@@ -4,7 +4,7 @@
   name 'LanguageMapping'
   description 'A mapping between two languages'
 
-  field :id, !types.ID do
+  field :id, !types.Int do
```
phyrog: Why this change?

eugenk: It's because language mappings don't have a slug. The `ID` was a bug. We query by the numerical `id` column.

phyrog: It might still make sense to use ID here. AFAIK Apollo uses ID fields as primary keys for caching.

eugenk: I think it's not an `ID` typed field, but an `id` named field:

> By default, InMemoryCache will attempt to use the commonly found primary keys of id and _id for the unique identifier if they exist along with __typename on an object.

https://www.apollographql.com/docs/react/basics/caching.html#normalization


This could mean that we should rename `locId` fields in the queries to `id`...

eugenk: ... but it seems we don't need to do that. See the next paragraphs of the above link:

> This “getter” behavior for unique identifiers can be configured manually via the dataIdFromObject option passed to the InMemoryCache constructor, so you can pick which field is used if some of your data follows unorthodox primary key conventions so it could be referenced by any query.
>
> For example, if you wanted to key off of the key field for all of your data, you could configure dataIdFromObject like so:
>
> ```
> const cache = new InMemoryCache({
>   dataIdFromObject: object => object.key || null
> });
> ```
> Note that Apollo Client doesn’t add the type name to the cache key when you specify a custom dataIdFromObject, so if your IDs are not unique across all objects, you might want to include the __typename in your dataIdFromObject.

PR #365, opened 2018-01-31
343 replace settings validator with dry validation

### LINKED ISSUE #343: Replace SettingsValidator with dry-validation

phyrog:
Closes #343.


config/initializers/settings_schema.rb
```diff
@@ -0,0 +1,85 @@
...
+
+# The Schema class to validate the Settings against
+class SettingsSchema < Dry::Validation::Schema
+  configure do |config|
+    config.messages_file = Rails.root.join(
+      'config/initializers/settings_validation_errors.yml'
+    )
+  end
+
+  def scheme?(list, value)
```
eugenk: Please prefix the name with `uri_has_`.
The same applies to the other URI-related predicates.

config/initializers/settings_schema.rb
```diff
@@ -0,0 +1,85 @@
...
+
+  def no_userinfo?(value)
+    URI(value).userinfo.nil?
+  end
+
+  def directory?(value)
+    File.directory?(value)
+  end
+
+  def create_directory?(dir)
```
eugenk: I would rather call this `create_directory` without the question mark because there is an action that is performed and not only a check.

spec/initializers/settings_schema_spec.rb
```diff
@@ -0,0 +1,138 @@
...
+        expect(subject.errors).to include(sneakers: ['must be an array'])
+      end
+
+      context 'workers' do
+        it 'is not a Numeric type' do
+          settings[:sneakers] = [workers: 'bad']
+          expect(subject.errors).to include(
+            sneakers: {0 => include(workers: ['must be an integer'])}
+          )
+        end
```
eugenk: What if this is a float?

config/initializers/settings_schema.rb
phyrog: Unfortunately this does not work with dry-validation. It must be suffixed with `?`.

spec/initializers/settings_schema_spec.rb
phyrog: Same error, I renamed the test to reflect that it needs to be an integer.

config/initializers/settings_schema.rb
phyrog: Fixed in 2dafd39

config/initializers/settings_schema.rb
eugenk: Ah, ok

config/initializers/settings_validation_errors.yml
```diff
@@ -0,0 +1,12 @@
+en:
+  errors:
+    scheme?: "has an invalid scheme (only %{list} are allowed)"
+    absolute?: "is not an absolute URL"
+    no_path?: "must not have a path"
+    no_query?: "must not have a query string"
+    no_fragment?: "must not have a fragment"
+    no_userinfo?: "must not have user info"
```
eugenk: Do you need to rename these as well? ;)

phyrog: Ah, yea...

PR #376, opened 2018-02-07
Don't mind me

phyrog:
Testing stickler CI


phyrog:
We just don't need the default configs in the repo because rubocop will use the defaults for all cops we don't configure otherwise. We need *just* `.rubocop_modified.yml`, so we can just name it `.rubocop.yml`. When a new rubocop version comes out, we only need to update our cop rules, and don't even have to touch the default files.

I think it makes upgrading easier, especially since most cop namespace changes won't affect the few changes we have compared to the defaults.

phyrog:
This `.rubocop.yml` and the four files we currently have should behave completely the same.

phyrog:
Side note: stickler currently uses rubocop 0.51, so also no Ruby 2.5 support.

PR #383, opened 2018-02-13
Fix Processes and Version Check in the System Tests

eugenk:
The system-tests fail when stopping all the applications because the multi runner's processes are already dead and the multi-runner itself can't kill them any more. This should fix that problem.

I'm currently changing the system-tests to start all applications in the production environment. The second commit fixes the version check for this purpose.


eugenk:
It is. I'm currently re-writing it.

eugenk:
Let's keep this open for some changes in the travis.yml I forgot about. (I remembered just now.)

eugenk:
I added the commit "Use switch_to_rspec_and_add_git_shell_tests branch for system-test." to show how the rspec system tests would work.

PR #385, opened 2018-02-14
Use bunny publish instead of sneakers

phyrog:
This should fix the problem where we can't send any messages to rabbitmq because of a timeout.

### DIFF
--- a/lib/hets_agent/invoker.rb
+++ b/lib/hets_agent/invoker.rb
@@ -14,14 +14,16 @@ module HetsAgent
     end
 
     def call
+      connection = Sneakers::CONFIG[:connection]
+      connection.start unless connection.open?
+      channel = connection.create_channel
+      exchange = channel.direct('sneakers', durable: true)
+
       request_collection.each do |request|
-        connection = Sneakers::CONFIG[:connection]
-        connection.start unless connection.open?
-        channel = connection.create_channel
-        exchange = channel.direct('sneakers', durable: true)
         exchange.publish(request.to_json, routing_key: WORKER_QUEUE_NAME)
-        connection.close
       end
+    ensure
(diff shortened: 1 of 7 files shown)


lib/hets_agent/invoker.rb
```diff
@@ -15,7 +15,12 @@ def initialize(request_collection)
 
     def call
       request_collection.each do |request|
-        Sneakers.publish(request.to_json, to_queue: WORKER_QUEUE_NAME)
+        conn = Sneakers::CONFIG[:connection]
+        conn.start unless conn.open?
+        ch = conn.create_channel
+        x = ch.direct('sneakers', durable: true)
+        x.publish(request.to_json, routing_key: WORKER_QUEUE_NAME)
```
phyrog: This is now published to the `sneakers` exchange instead of the queue directly.

lib/hets_agent/invoker.rb
```diff
@@ -15,7 +15,12 @@ def initialize(request_collection)
 
     def call
       request_collection.each do |request|
-        Sneakers.publish(request.to_json, to_queue: WORKER_QUEUE_NAME)
+        connection = Sneakers::CONFIG[:connection]
+        connection.start unless connection.open?
+        channel = connection.create_channel
+        exchange = channel.direct('sneakers', durable: true)
+        exchange.publish(request.to_json, routing_key: WORKER_QUEUE_NAME)
+        connection.close
```
eugenk: Please put `connection.close` in an `ensure`

eugenk:
I'm not sure that the style was broken and needed fixing...

phyrog:
This is way uglier than I hoped, but I couldn't find another way to fix that

PR #388, opened 2018-02-16
Add db:truncate and db:truncate:seed tasks

phyrog:
The problem with `db:recreate` is that all connections to the database are terminated. So `db:truncate` is useful when you want to keep all database connections and just empty all tables. So now you can do

```
# Terminal 1
rails invoker

# Terminal 2
rails db:truncate:seed
```

without having most applications restarted


phyrog:
Shouldn't be too hard with `DROP TABLE`. But doesn't a schema change require an application restart anyway?

eugenk:
I'm not sure about that in the development environment

phyrog:
feffe42 adds `db:recreate:tables` and `db:recreate:tables:seed`. For these to work, ontohub/ontohub-models#159 must be merged, or otherwise it will complain about duplicate functions.

phyrog:
Hmm, Bunny has [this method](http://reference.rubybunny.info/Bunny/Queue.html#purge-instance_method) on `Queue`, which means we would need to gather all queues to purge by hand

eugenk:
Should not be a big problem because we choose what new queues we introduce (i.e. there are not many of them).

phyrog:
Should `db:recreate` also purge the queues? Or just `db:recreate:tables` and `db:truncate`?

eugenk:
I think all of them should purge the queues. And the seeds should invoke that task, too (instead of using `db/seeds/001_clear_queue.rb`.

lib/tasks/rabbitmq.rake
```diff
@@ -0,0 +1,28 @@
...
+  hets\ >=\ 0.100.0
+  indexing
+  mailers
+  post_process_hets
+  process_commit
+)
+
+namespace :rabbitmq do
+  desc "Purge all rabbitmq queues"
+  task :purge do
```
eugenk: Not 100% sure, but I think that this needs the `:environment` because it uses `Settings`.

phyrog: Nope, I checked; none of them need `:environment`, not even for the database connection.

phyrog: Might have changed because of the whole `rails ...` instead of `rake ...`

eugenk:
I see just now that the seeds are failing. We need to override `RAILS_ENV=development` (Travis sets it to `test`). And we need to 
```
        # Set up RabbitMQ VirtualHost
        - sudo rabbitmqctl add_vhost ontohub_development
        - sudo rabbitmqctl set_permissions -p ontohub_development guest ".*" ".*" ".*"
```

in the `.travis.yml`

eugenk:
RSpec is failing because the vhost is not set up. Maybe we should skip `rabbitmq:purge` if `Rails.env.test?` is `true` since we don't connect to it anyway.

lib/tasks/rabbitmq.rake
```diff
@@ -0,0 +1,29 @@
...
+    next if Rails.env.test?
```
phyrog: Hm, like this?

eugenk: Yes

phyrog:
\*waiting for the next post-approval comment* :smile:

eugenk:
Tests are green. It's finally approved.

PR #399, opened 2018-02-26
Index seeds

phyrog:
This indexes the seeds when running `db:seed` et al.

### DIFF
--- a/db/seeds/020_organization_seeds.rb
+++ b/db/seeds/020_organization_seeds.rb
@@ -2,14 +2,14 @@
 
 Organization.
   create(name: 'seed-user-organization',
-      display_name: 'Seed User Organization',
-      description: 'All users that are created in the seeds').tap do |org|
+         display_name: 'Seed User Organization',
+         description: 'All users that are created in the seeds').tap do |org|
   IndexingJob.perform_later('class' => 'Organization', 'id' => org.id)
 end
 
 Organization.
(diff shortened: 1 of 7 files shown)


lib/tasks/db.rake
```diff
@@ -7,7 +7,7 @@ namespace :db do
     Rake::Task['invoker:start_all'].invoke
   end
 
-  task seed: ['rabbitmq:purge', 'chewy:reset'] do
+  task :seed do
```
eugenk: If you remove the prerequisites of the task, you can remove the whole task. Then, it is simply inherited from the models.

db/seeds/020_organization_seeds.rb
```diff
@@ -1,9 +1,15 @@
 # frozen_string_literal: true
 
-Organization.new(name: 'seed-user-organization',
-                 display_name: 'Seed User Organization',
-                 description: 'All users that are created in the seeds').save
+Organization.
+  new(name: 'seed-user-organization',
+      display_name: 'Seed User Organization',
+      description: 'All users that are created in the seeds').save.tap do |org|
```
eugenk: Please use `.create(...)` instead of `.new(...).save`. `.create(...)` is the same as `.new(...).save` except for the returned value: `save` returns a boolean while `create` returns the created object. So with `save`, the `tap` should not work as intended.

Issue #326, opened 2017-12-04
Provide basic search functionality


phyrog:
For org units and repositories (by name, description).

Possibly using [chewy](https://github.com/toptal/chewy).

PR #459, opened 2018-06-28
326 provide basic search functionality

DerProfessor:
Provides a basic search functionality described in #326


lib/search_resolver.rb
```diff
@@ -0,0 +1,86 @@
...
+        organizational_units: allOrganizationalUnits(result),
+        repositories: allRepositories(result)
+      )
+    )
+  end
+
+  private
+  def allOrganizationalUnits(result)
+    search_result = 0
+    result.each do |element|
```
phyrog: Check out [`Enumerable#count`](https://ruby-doc.org/core-1.9.3/Enumerable.html#method-i-count)

lib/search_resolver.rb
```diff
@@ -0,0 +1,86 @@
...
+        search_result+=1
+      end
+    end
+    search_result
+  end
+
+  private
+  def allRepositories(result)
+    search_result = 0
+    result.each do |element|
```
phyrog: Check out [`Enumerable#count`](https://ruby-doc.org/core-1.9.3/Enumerable.html#method-i-count)

lib/search_resolver.rb
```diff
@@ -0,0 +1,86 @@
...
+class SearchResolver
+  def call(root, arguments, _context)
+    query = root[:query]
+    categories = arguments[:categories]
+    if categories.blank?
+      indices = [::Index::RepositoryIndex::Repository,
+                  ::Index::OrganizationIndex::Organization,
+                  ::Index::UserIndex::User]
+    else
+      indices = categories.map do |category|
```
phyrog: Check out [`Enumerable#reduce`](https://ruby-doc.org/core-2.1.0/Enumerable.html#method-i-reduce):

```ruby
indices = categories.reduce do |indices, category|
  ...
    when 'repositories'
      indices + [::Index::RepositoryIndex::Repository
  ...
end
```

that way you don't need to flatten the array at the end, which might make it faster.

lib/search_resolver.rb
```diff
@@ -0,0 +1,86 @@
...
+          [::Index::OrganizationIndex::Organization, ::Index::UserIndex::User]
+        when 'repositories'
+          [::Index::RepositoryIndex::Repository]
+        else
+          []
+        end
+      end
+      indices = indices.flatten
+    end
+    result = indices.map do |index|
```
phyrog: Could be done using reduce as well, but I think here flatten is fine.

lib/search_resolver.rb
```diff
@@ -0,0 +1,84 @@
...
+                                fields: %i(
+                                  display_name
+                                  slug
+                                  name
+                                  description
+                                )}).entries
+    end.flatten
+  end
+
+  def create_indices(categories)
```
phyrog: `map_categories_to_indices` might be a better name

lib/search_resolver.rb
```diff
@@ -0,0 +1,84 @@
...
+                   ::Index::UserIndex::User]
+      when 'repositories'
+        indices + [::Index::RepositoryIndex::Repository]
+      else
+        indices + []
+      end
+    end
+  end
+
+  def all_organizational_units(result)
```
phyrog: `organizational_units_count` might be a better name

lib/search_resolver.rb
```diff
@@ -0,0 +1,84 @@
...
+    result.count do |element|
+      elem = element._data['_index']
+      elem == 'organization' || elem == 'user'
+    end
+  end
+
+  def all_repositories(result)
```
phyrog: `repositories_count` might be a better name

lib/search_resolver.rb
```diff
@@ -0,0 +1,84 @@
...
+    result.count do |element|
+      element._data['_index'] == 'repository'
+    end
+  end
+
+  def create_entries(result)
```
phyrog: `map_entries_to_models` might be a better name

lib/search_resolver.rb
```diff
@@ -0,0 +1,84 @@
...
+  def reduce_categories(categories)
+    categories.reduce([]) do |indices, category|
+      case category
+      when 'organizationalUnits'
+        indices + [::Index::OrganizationIndex::Organization,
+                   ::Index::UserIndex::User]
+      when 'repositories'
+        indices + [::Index::RepositoryIndex::Repository]
+      else
+        indices + []
```
phyrog: can be replaced by just `indices` as concatenating an empty list is a no-op.

